PR #69, opened 2020-09-09
Added PR, Issues, Feature Requests and Support templates

sahilaug:


joelrosario (changes requested):
@sahilaug Thanks, looks great!

- I made some cosmetic changes to the language in a couple of places.
- I don't think GitHub issues has an easy way to put in screenshots, so IMHO we should remove the screenshots section in the bug report template.

.github/PULL_REQUEST_TEMPLATE.MD
```diff
@@ -0,0 +1,37 @@
...
+this project (found in the CODE_OF_CONDUCT.md file).
+
+Also, please make sure you're familiar with and follow the instructions in the
+contributing guidelines (found in the CONTRIBUTING.md file).
+
+Please fill out the information below to expedite the review and (hopefully)
+merge of your pull request!
+-->
+
+<!-- What changes are being made? (What feature/bug is being fixed here?) -->
```
joelrosario, line 14: Shouldn't lines 14 and 16 be swapped? If so then the same is true of the why, how and checklist sections.

harikrishnan83 (approved):

PR #78, opened 2020-09-11
Test command changes

### LINKED ISSUE #64: Qontract json support in Commands
1. Contract Paths in JSON
2. Stubs in JSON
3. Host, Port and other config?

harikrishnan83:
**What**: Test Command should be able to take one or more contract files as arguments. When contract files are not provided, it should get the contracts to run as test in qontract.json

**Why**: To make Test Command consistent with Stub Command to enable running multiple Contracts as Tests.

**How**: Reuse the config file reading and contract install code

**Checklist**:

- [ ] Documentation added to the README.md OR link to PR on https://github.com/qontract/qontract-documentation N/A
- [x] Tests
- [x] Sonar Quality Gate

**Issue ID**:
Closes: #64 

I have also changed
1. junitReportDir option name
2. Removed workingDirectory option
3. Removed backwardCompatibility check in test command

### DIFF
--- a/junit5-support/src/main/kotlin/run/qontract/test/QontractJUnitSupport.kt
+++ b/junit5-support/src/main/kotlin/run/qontract/test/QontractJUnitSupport.kt
@@ -20,7 +20,7 @@ val pass = Unit
 open class QontractJUnitSupport {
     @TestFactory
     fun contractAsTest(): Collection<DynamicTest> {
-        val path = System.getProperty("path")
+        val contractPaths = System.getProperty("contractPaths")
         val givenWorkingDirectory = System.getProperty("workingDirectory")
         val givenConfigFile = System.getProperty("manifestFile")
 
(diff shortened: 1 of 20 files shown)


junit5-support/src/main/kotlin/run/qontract/test/QontractJUnitSupport.kt
```diff
@@ -18,30 +18,39 @@ import kotlin.system.exitProcess
 val pass = Unit
 
 open class QontractJUnitSupport {
+    companion object {
```
joelrosario, line 21: I prefer to avoid companion objects unless where necessary. Do you see any problem with moving these constants outside QontractJUnitSupport? They can sit by themselves without having to be in a class.

harikrishnan83: These are constants that are related only to the JunitSupport Class. We can move them after we see a pattern.

PR #133, opened 2020-11-27
Fixing: bundle command fails to pick qontracts & stub data in mono repo

### LINKED ISSUE #132: bundle command fails to pick qontracts & stub data in mono repo
**Description**
If a stub source in qontract.json is within the mono repo, the bundle command fails to pick the source contracts and stub data in the zip file.

**Steps to reproduce**
1. Create a qontract.json with stub source referring to contracts from within the same repo.
```json
{
  "sources": [
    {
      "provider": "git",
      "stub": [
        "../contractsDir/1.qontract"
      ]
    }
  ]
}
```

2. Run the bundle command: contract bundle

**Expected behavior**
The generated bundle.zip file should contain a base directory with the same name as the mono repo. The contracts and stub data should be placed within this base directory under the same path as it is placed in the actual mono repo.

**System Information:**
 - Qontract version: [0.20.1]

**Additional context**
n/a

deveshch:
Fixing: bundle command fails to pick contracts & stub data in mono repo (Issue#132)

**What**: 
Fixes #132 , makes sure the bundle command picks the contracts and stub data from within the mono repo, and places it under the base dir with the same name as the mono repo and under the same path as it is within the mono repo.

**Why**:
Currently, the bundle command is generating the zip file which is missing sources referring to the path within a mono repo.

**How**:
When the source belongs to a mono repo, ensuring contract file paths are picked correctly, relative to the mono repo root directory.

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

**Issue ID**:
Closes: #132

### DIFF
--- a/core/src/main/kotlin/run/qontract/core/git/SystemGit.kt
+++ b/core/src/main/kotlin/run/qontract/core/git/SystemGit.kt
@@ -11,9 +11,9 @@ class SystemGit(private val workingDirectory: String = ".", private val prefix: 
     override fun commit(): SystemGit = this.also { execute("git", "commit", "-m", "Updated contract") }
     override fun push(): SystemGit = this.also { execute("git", "push") }
     override fun pull(): SystemGit = this.also { execute("git", "pull") }
-    override fun resetHard(): SystemGit = this.also { execute("git",  "reset", "--hard", "HEAD") }
-    override fun resetMixed(): SystemGit = this.also { execute("git",  "reset", "--mixed", "HEAD") }
-    override fun mergeAbort(): SystemGit = this.also { execute("git",  "merge", "--aborg") }
+    override fun resetHard(): SystemGit = this.also { execute("git", "reset", "--hard", "HEAD") }
+    override fun resetMixed(): SystemGit = this.also { execute("git", "reset", "--mixed", "HEAD") }
+    override fun mergeAbort(): SystemGit = this.also { execute("git", "merge", "--aborg") }
     override fun checkout(branchName: String): SystemGit = this.also { execute("git", "checkout", branchName) }
     override fun merge(branchName: String): SystemGit = this.also { execute("git", "merge", branchName) }
     override fun clone(gitRepositoryURI: String, cloneDirectory: File): SystemGit = this.also { execute("git", "clone", gitRepositoryURI, cloneDirectory.absolutePath) }
@@ -21,7 +21,7 @@ class SystemGit(private val workingDirectory: String = ".", private val prefix: 
     override fun show(treeish: String, relativePath: String): String = execute("git", "show", "${treeish}:${relativePath}")
     override fun workingDirectoryIsGitRepo(): Boolean = try {
         execute("git", "rev-parse", "--is-inside-work-tree").trim() == "true"
-    } catch(e: Throwable) {
+    } catch (e: Throwable) {
         false.also {
             println("This must not be a git dir, got error ${e.javaClass.name}: ${exceptionCauseMessage(e)}")
         }
@@ -30,7 +30,7 @@ class SystemGit(private val workingDirectory: String = ".", private val prefix: 
     override fun getChangedFiles(): List<String> {
         val result = execute("git", "status", "--porcelain=1").trim()
 
-        if(result.isEmpty())
+        if (result.isEmpty())
             return emptyList()
 
         return result.lines().map { it.trim().split(" ", limit = 2)[1] }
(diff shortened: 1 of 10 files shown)


joelrosario (approved):

core/src/main/kotlin/run/qontract/core/git/SystemGit.kt
```diff
@@ -49,6 +49,11 @@ class SystemGit(private val workingDirectory: String = ".", private val prefix:
     }
 
     override fun inGitRootOf(contractPath: String): GitCommand = SystemGit(File(contractPath).parentFile.absolutePath)
+
+    override fun repoName(): String {
```
joelrosario: Might be better to do this using the directory instead of the url. Theoretically the file path can be different from the url, and the file path I believe is also easy to get at. I would like to get your thoughts on this.

core/src/main/kotlin/run/qontract/core/utilities/ContractSource.kt
```diff
@@ -58,14 +57,14 @@ data class GitRepo(val gitRepositoryURL: String, override val testContracts: Lis
     }
 }
 
-data class GitMonoRepo(override val testContracts: List<String>, override val stubContracts: List<String>) : ContractSource() {
+data class GitMonoRepo(override val testContracts: List<String>, override var stubContracts: List<String>) : ContractSource() {
```
joelrosario: I would like to avoid using mutable state as far as possible. It should be fairly easy to do. We can pair on this.

core/src/main/kotlin/run/qontract/core/utilities/Utilities.kt
```diff
@@ -253,29 +253,41 @@ fun contractFilePathsFrom(configFilePath: String, workingDirectory: String, sele
...
-                    defaultRepoDir.exists() && SystemGit(defaultRepoDir.path).workingDirectoryIsGitRepo() -> {
-                        println("Using local contracts")
-                        defaultRepoDir
-                    }
-                    else -> {
-                        val reposBaseDir = File(workingDirectory).resolve("repos")
-                        println("Couldn't find local contracts, cloning ${source.gitRepositoryURL} into ${reposBaseDir.path}")
-                        if(!reposBaseDir.exists())
-                            reposBaseDir.mkdirs()
+        println("Looking for contracts in local environment")
```
joelrosario: We may not need to check the local environment if we know the repo is a mono repo. We can pair on this.

core/src/main/kotlin/run/qontract/core/git/SystemGit.kt
deveshch: removed reponame, making use of gitRoot now.

core/src/main/kotlin/run/qontract/core/utilities/ContractSource.kt
deveshch: refactored to add a new method to the ContractSource - loadContracts, which keeps the ContractSource immutable.

core/src/main/kotlin/run/qontract/core/utilities/Utilities.kt
deveshch: Done.

core/src/main/kotlin/run/qontract/core/git/SystemGit.kt
```diff
@@ -11,17 +11,17 @@ class SystemGit(private val workingDirectory: String = ".", private val prefix:
     override fun commit(): SystemGit = this.also { execute("git", "commit", "-m", "Updated contract") }
     override fun push(): SystemGit = this.also { execute("git", "push") }
     override fun pull(): SystemGit = this.also { execute("git", "pull") }
-    override fun resetHard(): SystemGit = this.also { execute("git",  "reset", "--hard", "HEAD") }
-    override fun resetMixed(): SystemGit = this.also { execute("git",  "reset", "--mixed", "HEAD") }
-    override fun mergeAbort(): SystemGit = this.also { execute("git",  "merge", "--aborg") }
+    override fun resetHard(): SystemGit = this.also { execute("git", "reset", "--hard", "HEAD") }
+    override fun resetMixed(): SystemGit = this.also { execute("git", "reset", "--mixed", "HEAD") }
+    override fun mergeAbort(): SystemGit = this.also { execute("git", "merge", "--aborg") }
```
joelrosario, line 16: Can you make this abort instead of aborg. This may have been a pre existing bug, even before this PR.

joelrosario (approved):

PR #330, opened 2021-09-20
externalising stub response command

harikrishnan83:
**What**:

In Service Virtualisation mode, Stub response at times needs to be based on parameters in the request.

**Why**:

Example: If the Component Under Test makes an API call which involves a collection of entities with ids (DB primary keys which are generated by Component Under Test and so cannot be predicted ahead of time to include in the stub json), then it will depend on the stub response having those very same ids. If Stub response includes random (specmatic generated) / static ids (stub json) then it can lead to Component Under Test throwing errors because it cannot find those ids in its DB.

**How**:
Specmatic should allow delegating the responsibility of parsing the request and generating a stub response based on values in the request to an external command which is authored by the Users.
* Users need to add "externalisedResponseCommand" field as part of http-response in stub json and set it to any command that is available on the machine where specmatic is running.
* When present this command is called by Specmatic with the request json in an environment variable named "SPECMATIC_REQUEST". Specmatic expects the command to return only the response json. The command can be any script (Shell Scripts, Ruby, Python or any other language scripts, etc.) that is executable in the environment where Specmatic is running. Example: [Shell Script](https://github.com/znsio/specmatic/pull/330/files#diff-6ce4f5b816ed054c6f6e90da4d96c7ccbb69fe5877925e3aa7c0f1536b29aabe)
* Specmatic then validates the response against the associated specification and returns it

Pending

- [x] Command not found
- [x] Error Executing Command 
- [x] Cleaner approach to serialise request to a single line JSON string - Moving to Kotlinx Serialisation
- [x] Pass request JSON as an Environment Variable
- [ ] Test with large requests and responses

**Checklist**:

- [ ] Documentation added to the README.md OR link to PR on https://github.com/qontract/qontract-documentation - Pending
- [x] Tests
- [x] Sonar Quality Gate

**Other Changes**:
* Removing jcenter references - [https://blog.gradle.org/jcenter-shutdown](https://blog.gradle.org/jcenter-shutdown)


joelrosario:
Closing, merged it manually.

PR #518, opened 2022-10-01
reading openapi string regular expression pattern

harikrishnan83:
**What**:

String Pattern should consider the regular expression in OpenAPI while matching and generating values 

**Why**:

* Contract Tests
  * Request - Values generated should be according to regular expression since application may be expecting it
  * Response - Verification of Strings against regular expression makes tests even more thorough
* Stubs
  * Request - Verification of Strings against regular expression makes the stub behave exactly as per spec
  * Response - Generating values as per regular expression emulates the Provider even better

**How**:

* Matching leverage Kotlin libraries
* Generating is done with [generex](https://github.com/mifmif/Generex)

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -42,6 +42,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.4'
 
     implementation 'io.swagger.parser.v3:swagger-parser:2.1.3'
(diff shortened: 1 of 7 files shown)


harikrishnan83:
Closing this because the PR is quite old and we need to chose another approach to generate stub responses based on regex.

PR #581, opened 2023-01-13
Ensure that DateTimePattern is stringified as its pattern

### LINKED ISSUE #580: Using date-time in a query param does not generate a timestamp
**Description**
My API takes two timestamps as query parameters.

**Steps to reproduce**
1. Create a `api.yaml` with 
```yaml
openapi: 3.0.1
info:
  title: New Feature
  version: "1"
paths:
  /customer/name:
    get:
      summary: New scenario
      parameters:
      - name: timestamp
        in: query
        schema:
          type: string
          format: date-time
        example: '2023-01-12T12:53:10.16048'
      responses:
        "200":
          description: New scenario
          content:
            text/plain:
              schema:
                type: string
```
2. Run this command: `specmatic samples api.yaml`

**Expected behavior**
That specmatic creates example requests with timestamps. 

Instead it generates this request (where instead of a timestamp, it is the object ID):
```json
{
    "requestTime": "2023-1-13 0:45:23.750",
    "http-request": {
        "path": "/customer/name",
        "method": "GET",
        "query": {
            "timestamp": "in.specmatic.core.pattern.DateTimePattern@5fef2aac"
        },
        "headers": {
            "Accept-Charset": "UTF-8",
            "Accept": "*/*"
        },
        "body": ""
    }
}
```

**System Information:**
 - Specmatic version: 0.59.0

**Additional context**
It behaves similarly when trying to run it with contract tests. It seems to be some issue with how date-time is serialized.

androa:
**What**: DateTimePattern is casted to string by it's own JVM identifier, but it should be the `pattern`.

**Why**: Several places seem to be replacing the `(datetime)` pattern with an actually value

**Checklist**:

- [x] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation N/A
- [x] Tests
- [x] Sonar Quality Gate

**Issue ID**:
Closes: #580 

I've been trying to find a nice place to test this, but it seems to fall a bit between the cracks of the test-suite. Any pointers on where it would be appropriate to add tests would be very appreciated.

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/pattern/DateTimePattern.kt
+++ b/core/src/main/kotlin/in/specmatic/core/pattern/DateTimePattern.kt
@@ -47,6 +47,8 @@ object DateTimePattern : Pattern, ScalarType {
     override val typeName: String = "datetime"
 
     override val pattern = "(datetime)"
+
+    override fun toString() = pattern
 }
 
 fun currentDate() = StringValue(LocalDateTime.now().format(DateTimeFormatter.ISO_DATE_TIME))
--- a/core/src/test/kotlin/in/specmatic/core/URLMatcherTest.kt
+++ b/core/src/test/kotlin/in/specmatic/core/URLMatcherTest.kt
@@ -237,4 +237,10 @@ internal class URLMatcherTest {
         val result = matcher.matches(URI("/"), mapOf("name" to "Archie"), Resolver())
         assertThat(result.isSuccess()).isFalse()
     }
+
+    @Test
+    fun `should stringify date time query param to date time pattern`() {
+        val urlMatcher = URLMatcher(mapOf("before" to DateTimePattern), pathToPattern("/pets"), "/pets")
+        assertThat(urlMatcher.toString()).isEqualTo("/pets?before=(datetime)")
+    }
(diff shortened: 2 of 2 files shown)


harikrishnan83:
@androa Thanks for your Pull Request. I think you should be able to add a test like this to [UrlMatcherTest.kt](https://github.com/znsio/specmatic/blob/main/core/src/test/kotlin/in/specmatic/core/URLMatcherTest.kt)

```kotlin
@Test
fun `should stringify date time query param to date time pattern`() {
    val urlMatcher = URLMatcher(mapOf("before" to DateTimePattern), pathToPattern("/pets"), "/pets")
    assertThat(urlMatcher).isEqualTo("/pets?before=(datetime)")
}
```

One you have added this test and once the build is passing we will test the changes and merge the PR. Please let us know if you have any questions on this.

androa:
Thank you @harikrishnan83 for the help! Added the test case (it needed to be converted to string to satisfy the equal check). Works perfect.

Also tested it with the repro YAML from the issue and it generates samples and stubbing perfectly now.

harikrishnan83 (approved):
Looks good to me

PR #586, opened 2023-01-19
Support oneOf multiple sub-schemas

westse:
- Previously only supported the first sub-schema under oneOf (plus a nullable sub-schema)

**What**: Adds support for selecting any one of multiple sub-schemas under oneOf. Currently only the first found sub-schema is supported.

**Why**: Fixes issue #585 

**How**: Made existing oneOf support more complete in OpenApiSpecification.kt

**Checklist**:

- [ ] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation
- [X] Tests 
- [ ] Sonar Quality Gate N/A

**Issue ID**:
Closes: #585

Issue #585, opened 2023-01-19
oneOf only supports first sub-schema


westse:
**Description**
Only the first sub-schema of oneOf is supported. Usage of any other sub-schema results in an error.

**Steps to reproduce**
1. See unit test in pull request #586 

**Expected behavior**
Any one of multiple sub-schemas of oneOf should be supported

**System Information:**
 - OS & version: Ubuntu
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

**Additional context**
Also blocks resolution of another bug related to a dictionary with values with oneOf schema #587

sergeyklay:
I think this is related too: https://github.com/znsio/specmatic/pull/588#issuecomment-1399256057

harikrishnan83:
@westse as mentioned by @joelrosario we had to look at this failing scenario 437e242be4650519df78d1445a609b1b0e2bae6a and we have subsequently fixed the same with 24340d8ee6b55b98dc30cfbc378596616460f728

PR #588, opened 2023-01-19
Support dictionary with allOf, oneOf values

### LINKED ISSUE #587: Dictionary of oneOf, allOf produces NullPointerException
**Description**
Dictionary of oneOf or allOf should be supported

**Steps to reproduce**
1. See unit test in pull request #588 
2. Produces `java.lang.NullPointerException: valueSchema.types must not be null`

**Expected behavior**
Dictionary of oneOf or allOf should be supported with no error

**System Information:**
 - OS & version: Ubuntu
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

**Additional context**
Related to #585 which must be fixed before this issue can be fully fixed.

westse:
**What**: Adds support for dictionary of oneOf or allOf value schemas. Without this fixed a NullPointerException is produced. NOTE: #585 must be fully fixed ( PR #586 merged) before this can be fully fixed.

**Why**: Fixes issue #587 

**How**: Updated existing support in `OpenApiSpecification.toDictionaryPattern()`

**Checklist**:

- [ X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #587 

NOTE: #585 must be fully fixed ( PR #586 merged) before this can be fully fixed.

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
@@ -779,7 +779,7 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
         schema: Schema<*>, typeStack: List<String>, patternName: String
     ): DictionaryPattern {
         val valueSchema = schema.additionalProperties as Schema<Any>
-        val valueSchemaTypeName = if (valueSchema.`$ref` == null) valueSchema.types.first() else valueSchema.`$ref`
+        val valueSchemaTypeName = valueSchema.`$ref` ?: valueSchema.types?.first() ?: "";
         return DictionaryPattern(
             StringPattern(), toSpecmaticPattern(valueSchema, typeStack, valueSchemaTypeName, false)
         )
--- a/core/src/test/kotlin/in/specmatic/conversions/OpenApiSpecificationTest.kt
+++ b/core/src/test/kotlin/in/specmatic/conversions/OpenApiSpecificationTest.kt
@@ -3494,6 +3494,69 @@ components:
         assertThat(stub.response.status).isEqualTo(200)
 
     }
+  @Test
+  fun `support dictionary object type with composed oneOf value`() {
+    val openAPI =
+      """
+---
+openapi: 3.0.1
+info:
+  title: API
+  version: 1
+paths:
+  /data:
+    post:
+      summary: API
+      parameters: []
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                myValues:
+                  type: object
+                  additionalProperties:
+                    oneOf:
+                    - ${"$"}ref: Person
+                    - ${"$"}ref: Alien
+      responses:
+        200:
+          description: API
+components:
+  schemas:
+    Person:
+      type: object
+      properties:
+        name:
+          type: string
+    Alien:
+      type: object
+      properties:
+        moniker:
+          type: string
+        homePlanet:
+          type: string
+""".trimIndent()
+
+    println(openAPI)
+    val feature = OpenApiSpecification.fromYAML(openAPI, "").toFeature()
+
+    val request = HttpRequest(
+      "POST",
+      "/data",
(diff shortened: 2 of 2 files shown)


sergeyklay:
@westse Could you please add test for `allOf` too. Just faced with the issue that specmatic doesn't support inheritance via `allOf`, and I'm interested in getting this fixed.

For example, the server I'm testing might return a response in the following format:
```json
{
  "errors": [
    {
      "title": "validation.required",
      "status": 422,
      "source": {
        "pointer": "/data/attributes/url"
      }
    }
  ]
}
```

In fact, this is not the only kind of error that can be returned. For example, in some situations the server may return the following response body:

```json
{
  "errors": [
    {
      "title": "A short, human-readable summary",
      "status": 400
    }
  ]
}
```

Or even:
```diff
{
  "errors": [
    {
      "title": "A short, human-readable summary",
-     "status": 400
+     "status": 400,
+     "source": {
+       "pointer": "/data/attributes/attribute_name"
+     }
    }
  ]
}
```

Here the `source` attribute is optional.

As you can see, these errors have a number of things in common. Thus, my naive attempt was to reduce the repetitive code in the OpenAPI specification through inherited:

```yaml
components:
  schemas:
    BasicErrorModel:
      type: object
      required:
        - errors
      properties:
        errors:
          type: array
          items:
            type: object
            required:
              - title
              - status
            properties:
              # This attribute will be replaced after merging
              title:
                type: string
                example: A short, human-readable summary
              # This attribute will be replaced after merging
              status:
                type: integer
                example: 400
              source:
                type: object
                description: An object containing references to the primary source of the error.
                properties:
                  pointer:
                    type: string
                    description: |
                      A JSON Pointer to the value in the request document that caused the error.
                      This MUST point to a value in the request document that exists; if it doesn't, the client SHOULD
                      simply ignore the pointer.
                    example: /data/attributes/some_attribute

  responses:
    '400':
      description: The request could not be understood by the server due to malformed syntax.
      content:
        application/vnd.api+json:
          schema:
            allOf: # Combines the BasicErrorModel and the inline model
              - $ref: '#/components/schemas/BasicErrorModel'
              - type: object
                properties:
                  errors:
                    type: array
                    items:
                      type: object
                      properties:
                        status:
                          type: integer
                          example: 400

    '422':
      description: The server was unable to process the contained instructions.
      content:
        application/vnd.api+json:
          schema:
            allOf: # Combines the BasicErrorModel and the inline model
              - $ref: '#/components/schemas/BasicErrorModel'
              - type: object
                properties:
                  errors:
                    type: array
                    items:
                      type: object
                      properties:
                        status:
                          type: integer
                          example: 422
```

However, after running the contract tests, I found that specmatic simply doesn't use inheritance (it does not combine everything in the `allOf` section). When I run a contract test, I get the following log:

```
    	API: POST /the/server/url/here -> 422

    	  >> RESPONSE.BODY.errors[0].source

    	     Key named source in the response was not in the contract

    	  >> RESPONSE.BODY.errors[0].source

    	     Key named source in the response was not in the contract
```

@harikrishnan83 I'm not sure if this should follow in a separate issue. Feel free to move this to a separate issue if necessary.

westse:
@sergeyklay I suggest opening a separate issue with those details. I believe it has to do with the fact that currently specmatic will only include required allOf properties, and not optional ones (which your `source` property is). Perhaps @harikrishnan83 can comment on the reason for this limitation. The relevant code [appears to be here](https://github.com/znsio/specmatic/blob/b34f8408768209ce1ec229a4b6a178421bfab3d4/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt#L569).

Note I have just opened a related issue #593 (and pull request #594) to add discriminator support which would be required [to support true inheritance](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism).

harikrishnan83:
@westse we (@joelrosario and I) have fully merge #586, please go ahead with necessary changes for this issue in the related PR

harikrishnan83:
@sergeyklay Thanks for reporting the issues you saw with allOf. I have created a separate issue #601 with the details you have shared. cc @joelrosario

westse:
@harikrishnan83 @joelrosario I have rebased on master and should be good to go.

harikrishnan83 (approved):
Looks good to me. Thanks a lot @westse. cc @joelrosario

Issue #583, opened 2023-01-13
StackOverflowError for components with indirect circular reference


westse:
**Description**
While direct circular references between OpenApi components appears to be supported, indirect circular references cause specmatic to fail with a StackOverflowError.

**Steps to reproduce**
1. Create a file `circular-reference.yaml` with the following content

```
openapi: 3.0.1
info:
  title: Demonstrate circular reference bug in specmatic
  version: "1.0"
paths:
  /demo/circular-reference:
    get:
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Root'

components:
  schemas:

    Root:
      type: object
      properties:
        # A direct cycle works
        direct-cycle:
          $ref: '#/components/schemas/Root'

        # An indirect cycle via an intermediate node does NOT work
        intermediate-node:
          $ref: '#/components/schemas/IntermediateNode'


    IntermediateNode:
      type: object
      properties:
        # Completes an indirect cycle back to the root
        indirect-cycle:
          $ref: '#/components/schemas/Root'
```

2. Run this command: `java -jar specmatic.jar stub circular-reference.yaml`
3. Observe error: `Could not load circular-reference.yaml: Exception class=java.lang.StackOverflowError`
4. For full stack trace, run again with `--debug` option.

**Expected behavior**
Indirect circular references between components should be supported, similar to direct circular references.

**System Information:**
 - OS & version: Ubuntu 18.04
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

PR #592, opened 2023-01-20
Support component cycles in stub generation

### LINKED ISSUE #591: component cycle stub generation StackOverflowError
**Description**
Generating stubs and tests for OpenApi spec containing component cycles causes StackOverflowError.

**Steps to reproduce**
1. See unit tests in pull request #608 

**Expected behavior**
Stub generation and contract testing without error

**Screenshots**
n/a

**System Information:**
 - OS & version: Ubuntu
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

**Additional context**

westse:
**What**: Fixes #591 and StackOverflowError produced when generating stubs for spec containing component cycles.

**Why**: Avoids serious StackOverflowError and better supports the OpenApi specification

**How**: Resolver keeps a stack of patterns visited in order to avoid infinite processing cycle.
- Otherwise StackOverflowError thrown
- Uses correct pattern for the referred-to component that would introduce a cycle (previously would use the referring component's pattern). See OpenApiSpecification.toSpecmaticPattern()
- Test enhanced to test stub generation for direct cycle, indirect cycles through another component, and cycles through arrays.

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #591 

Related to #583 , but this issue deals with stub generation logic.

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Resolver.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Resolver.kt
@@ -82,7 +82,7 @@ data class Resolver(
             val subStack = cyclePreventionStack.subList(index, cyclePreventionStack.size)
             if (subStack.stream().noneMatch { nullable(it) }) {
                 // Terminate what would otherwise be an infinite cycle.
-                throw PatternCycleException("Invalid pattern cycle", cyclePreventionStack.plus(pattern))
+                throw ContractException("Invalid pattern cycle: ${cyclePreventionStack.plus(pattern)}", isCycle = true)
             }
         }
     }
--- a/core/src/main/kotlin/in/specmatic/core/Resolver.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Resolver.kt
@@ -77,7 +77,7 @@ data class Resolver(
 
     private fun validateNoCycle(pattern: Pattern) {
         val index = cyclePreventionStack.indexOf(pattern)
-        if (index > 0) {
+        if (index >= 0) {
             // Nullables are allowed since they will eventually terminate any cycle (only consider substack with cycle)
             val subStack = cyclePreventionStack.subList(index, cyclePreventionStack.size)
             if (subStack.stream().noneMatch { nullable(it) }) {
--- a/core/src/main/kotlin/in/specmatic/core/Resolver.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Resolver.kt
@@ -13,7 +13,8 @@ data class Resolver(
     val context: Map<String, String> = emptyMap(),
     val mismatchMessages: MismatchMessages = DefaultMismatchMessages,
     val isNegative: Boolean = false,
-    val generativeTestingEnabled: Boolean = false
+    val generativeTestingEnabled: Boolean = false,
+    val cyclePreventionStack: MutableList<Pattern> = mutableListOf(),
 ) {
     constructor(facts: Map<String, Value> = emptyMap(), mockMode: Boolean = false, newPatterns: Map<String, Pattern> = emptyMap()) : this(CheckFacts(facts), mockMode, newPatterns)
     constructor() : this(emptyMap(), false)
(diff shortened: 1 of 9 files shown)


westse:
Update: I have removed `Pattern.isLeaf()` (and all implementations) from this pull-request because I realized it was not the really the optimization I thought. Hopefully that makes your review simpler. The core change is in `Resolver.generate()`

harikrishnan83:
> Update: I have removed `Pattern.isLeaf()` (and all implementations) from this pull-request because I realized it was not the really the optimization I thought. Hopefully that makes your review simpler. The core change is in `Resolver.generate()`

@westse thanks, we will take a look

core/src/main/kotlin/in/specmatic/core/Resolver.kt
```diff
@@ -70,8 +72,18 @@ data class Resolver(
         }
 
     fun generate(factKey: String, pattern: Pattern): Value {
```
harikrishnan83, line 90: @westse Thanks for covering this in ```OpenApiKtTest.kt```. However, can we also cover this method with a unit test in ```ResolverKtTest```? As you may have noticed already our ```OpenApiKtTest``` is becoming quite large and a little hard to manage all functionality at that level. Also since Resolver is not particularly associated with openapi it may be easier / appropriate to test it independently anyway.

joelrosario (changes requested):
Hi @westse 

We understand the problem highlighted in the issue #591. Specmatic should certainly not be throwing a StackOverflowException.

However, given that the cyclical reference in your example / test in this PR is not nullable, the resolver cannot return a null to terminate the cycle, as that would not be contract-valid.

Can you update the PR to throw a ContractException exception, indicating the type that was found in the typeStack in the resolver, which has a cyclical reference.

Note: The API Specification can be updated to make the cyclical reference a nullable value. Nullable values are represented by AnyPattern, containing NullPattern as one of the candidate values. When AnyPattern generates a value, it picks randomly between the candidate values. At some point, it would pick NullPattern and generate a value, thus breaking the cycle.

westse:
@joelrosario Thanks for the clarification. I have updated this PR per your suggestions.

westse:
Note I just opened #608 for a more thorough eradication of StackOverflowError, although this PR probably should be merged first.

core/src/main/kotlin/in/specmatic/core/pattern/JSONObjectPattern.kt
```diff
@@ -102,9 +103,27 @@ data class JSONObjectPattern(override val pattern: Map<String, Pattern> = emptyM
...
+    val optionalProps = jsonPattern.keys.filter { isOptional(it) }.map { withoutOptionality(it) }
+
+    return jsonPattern
+        .mapKeys { entry -> withoutOptionality(entry.key) }
+        .mapValues { (key, pattern) ->
+            attempt(breadCrumb = key) {
+                try {
+                    Optional.of(resolverWithNullType.generate(key, pattern))
+                } catch (e: PatternCycleException) {
+                    if (optionalProps.contains(key)) {
```
harikrishnan83: @westse In this location, the optionality of the key itself may be relevant to break the cycle. What we should checking for is if the nullability of the value. We will push this fix to your branch shortly so that you can also take a look at the changes and share your thoughts. Hope that is okay. cc @joelrosario

core/src/main/kotlin/in/specmatic/core/Resolver.kt
```diff
@@ -69,9 +70,35 @@ data class Resolver(
...
+    }
+
     fun generate(factKey: String, pattern: Pattern): Value {
-        if (!factStore.has(factKey))
-            return pattern.generate(this)
+        if (!factStore.has(factKey)) {
+
+            validateNoCycle(pattern)
+
+            cyclePreventionStack.add(pattern)
```
harikrishnan83, line 95: @westse We notice that cyclePreventionStack is a multable list which is an instance variable of Resolver, which thereby makes Resolver itself mutable. Because of the nature of how we pass resolver around, it is extremely critical to keep the resolve immutable. We will make the necessary changes (passing the type stack as method parameter thereby not requiring instance variable) and push it your branch. cc @westse

harikrishnan83 (commented):
@westse Thanks again.

@joelrosario and I have taken a look at this PR and we have left a couple of comments. We need to make two changes to this PR before we can merge it.
* Logical problem in JSONObjectPattern where we are handling PatternCycleException
* Resolver by design has to be immutable. The cyclePreventionStack by virtue of being a mutable collection also makes Resolver mutable, which needs to be addressed.
As mentioned in the comments, we will make these changes and push them to your branch.

core/src/main/kotlin/in/specmatic/core/Resolver.kt
westse: Agreed about the immutability (I was just trying to not be too invasive :smile: ).

core/src/main/kotlin/in/specmatic/core/pattern/JSONObjectPattern.kt
westse: Sorry, I'm not sure I follow. I may be missing a concept, but let me describe the problem I'm addressing and this current solution. (Also, I apologize that somehow I originally missed including a unit test to expose the limitation being addressed. I have since pushed a test along with a necessary refactor.)

Stepping back, this PR is meant to have specmatic behave well in the face of cycles in patterns (e.g. A->B->A). Per @joelrosario 's feedback, cycles in required, non-nullable patterns should result in an exception. However cycles in nullable patterns (both required and optional) can be allowed because we can rely on AnyPattern.generate()'s behavior to eventually randomly select null and terminate the cycle. (That is what the Resolver cyclePreventionStack was for)

However, there is another case: cycles where one or more non-nullable properties are optional (ie. do no appear in the required list). Rather than failing due to a cycle it seems preferrable (and contract-valid) to instead omit the optional property in the generated stub. That is what this logic is meant to accomplish.

core/src/main/kotlin/in/specmatic/core/Resolver.kt
westse: That makes sense - but since you plan to refactor the cyclePreventionStack anyway perhaps you can add it then?

westse:
Closing this in favor of #608 which tests and fixes not only this issue but also StackOverFlowError in polymorphic scenarios. Note that the recommendation for keeping Resolver immutable has been addressed in #608 as well.

PR #598, opened 2023-01-26
Support Xml response with type: string

### LINKED ISSUE #597: Xml response with type: string produces ContractException
**Description**
Xml response with type: string produces `ContractException(errorMessage=Node not recognized as XML type: string, breadCrumb=, exceptionCause=null, scenario=null)`

**Steps to reproduce**
1. See unit test of pull request #598 

**Expected behavior**
Support Xml response with type: string

**Screenshots**
n/a

**System Information:**
 - OS & version: Ubuntu
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

**Additional context**

westse:
**What**: Avoids `ContractException(errorMessage=Node not recognized as XML type: string, breadCrumb=, exceptionCause=null, scenario=null)`

**Why**: Fixes #597 .

**How**: Uses string pattern when type: string

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #597

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
@@ -628,6 +628,9 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
     }
 
     private fun toXMLPattern(mediaType: MediaType): Pattern {
+        if (mediaType.schema is StringSchema)
+            return toSpecmaticPattern(mediaType.schema, emptyList())
+
         return toXMLPattern(mediaType.schema, typeStack = emptyList())
     }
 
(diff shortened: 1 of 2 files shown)


joelrosario:
@westse Thanks for reporting the issue. Based on the example in the test case in your PR, we noticed that you are trying to represent an XML response without specifying any XML schema for the response. This makes for a very loose API specification. As you may already know, Openapi [does have support for XML](https://swagger.io/docs/specification/data-models/representing-xml/). The exception you are seeing is by design, as we would like to encourage people to be more specific about how they author API Specifications.

That said, we'd like to understand your need for a lenient specification, as with this we lose the advantage of contract verification.

westse:
Good point, closing.

PR #600, opened 2023-01-26
Support query parameter free-form dictionary

### LINKED ISSUE #599: Query param of free-form dictionary produces URISyntaxException
**Description**
Query param of free-form dictionary produces `URISyntaxException: Illegal character in query at index...`

**Steps to reproduce**
1. See pull request  #600 

**Expected behavior**
No exception and free-form dictionary should be supported per OAS 3.x

**Screenshots**
n/a

**System Information:**
 - OS & version: n/a
 - Browser & version: n/a
 - Specmatic version: 0.59.0
 - JDK version: 11

**Additional context**

westse:
- Also with enum values

**What**: Supports query param of free-form dictionary without error.

**Why**: Fixes #599 

**How**: Updated `toSpecmaticPath()`

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #599


core/src/test/kotlin/in/specmatic/conversions/OpenApiSpecificationTest.kt
```diff
@@ -3487,7 +3487,86 @@ paths:
...
+            type: string
+      responses:
+        200:
+          description: API
+""".trimIndent()
+
+    println(openAPI)
+    val feature = OpenApiSpecification.fromYAML(openAPI, "").toFeature()
+
+    val request = HttpRequest("GET", "/data")
```
harikrishnan83: @westse can you please update the URL to include a sample query parameter also?

Not sure how to test this
* I extracted the OpenAPI yaml in this test to a file - ```test.yaml```
* Then I ran specmatic stub server with this yaml file - ```shell specmatic stub test.yaml```
* I also pasted the contents of ```test.yaml``` into editor.swagger.io and generated a curl command from there and ran it agains the stub server above and it does not seem to be working

```shell
curl -X 'GET' \
  'http://localhost:9000/data?additionalProp1=string&additionalProp2=string&additionalProp3=string' \
  -H 'accept: */*'
In scenario "GET /data?(string)=(string). Response: API"
API: GET /data -> 200

  >> REQUEST.QUERY-PARAMS.additionalProp1

     Query param named additionalProp1 in the request was not in the contract

  >> REQUEST.QUERY-PARAMS.additionalProp2

     Query param named additionalProp2 in the request was not in the contract

  >> REQUEST.QUERY-PARAMS.additionalProp3

     Query param named additionalProp3 in the request was not in the contract
```

Am I missing something? Can you please share an example of URL with free form dictionary so that I can test this pull request from both a stubbing and testing point of view?

harikrishnan83 (dismissed):

core/src/test/kotlin/in/specmatic/conversions/OpenApiSpecificationTest.kt
westse: @harikrishnan83 good catch. Both the test and implementation were incomplete, but should be fixed now. Note that I'm not confident in the implementation, but it works until there is a better solution?

I missed that because for my purposes I just needed specmatic to load my openapi.yaml without abruptly exiting, but do not yet need the stubbed endpoint to actually work since my contract testing does not exercise all of the endpoints in stubbed service. It seems like specmatic should be more resilient in this regard, stubbing what it can and only log errors about what it cannot. Is it worth creating an issue for this?

harikrishnan83: @westse sorry we were able to get back to you earlier on this PR. We are looking into it now. Thanks. cc @joelrosario

joelrosario (approved):
Hi @westse, apologies for the long delay on this PR.

We have at long last reviewed the OpenAPI documentation on the subject.

The implementation in this PR ignores keys that are not in the specification when additionalProperties is used in the definition of a query parameter. Please correct us if we're wrong. Instead, the datatypes of the additional parameters should be validated.

The code has also undergone several changes since you raised it. We have added commits to fix the above issue, added tests, and updated the PR in line with recent development on the codebase.

Approving and merging this PR. Thanks Steve!

PR #615, opened 2023-02-28
Chain ContractException with its cause

### LINKED ISSUE #614: ContractException cause lost when stack printed
**Description**
ContractException's constructor accepts an `exceptionCause` param for other reasons, but does not pass it to the super constructor per standard practice for chaining exceptions to their cause. This makes it difficult to find the root cause of many issues and contribute to the project

**Steps to reproduce**
1. Ensure an exception (e.g. `throw IllegalStateException("@@@ Intentional exception @@")` is thrown in a core location (such as first line of [DefinedPattern.newBasedOn(Row, Resolver)](https://github.com/znsio/specmatic/blob/81304aaa3894a33fd840fa127584824426744142/core/src/main/kotlin/in/specmatic/core/pattern/DeferredPattern.kt#L24))
2. Run `specmatic test`
3. Observe the printed stacktrace does not include a stack for the root cause.

**Expected behavior**
Exceptions should be chained so root cause analysis can be performed.

**System Information:**
 - OS & version: n/a
 - Browser & version: n/a
 - Specmatic version: 0.60.0
 - JDK version: n/a

westse:
- Otherwise traceability into the root causes of exceptions is lost.

**What**: Chains `ContractException` with its cause.

**Why**: To fix #614 and make it possible to perform root cause analysis on exceptions.

**How**: Follows standard practice of chaining an exception with its cause by passing the exception cause to the super class.

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #614

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Scenario.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Scenario.kt
@@ -207,7 +207,8 @@ data class Scenario(
             )
                 code()
         } catch (e: Throwable) {
-            throw ContractException("Couldn't match state values. Expected $expectedValue in key $key, actual value is $actualValue")
+            throw ContractException("Couldn't match state values. Expected $expectedValue in key $key" +
+                ", actual value is $actualValue", exceptionCause = e)
         }
     }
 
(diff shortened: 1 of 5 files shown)


joelrosario (approved):
Looks good to me

harikrishnan83:
Thanks a lot @westse

PR #617, opened 2023-03-04
Allow invalid request examples

joelrosario:
**What**:

Examples can now be marked WIP. A WIP example is similar to a WIP scenario, in that if the example fails, it gets logged but is not considered a failure but Specmatic. JUnit will ignore the failure, the return value on the command-line will be 0 (if the failing test from the WIP example is the only failure) and the failure will not break any build.

**Why**:

There are times when we wish to add an example that should trigger a 400 but instead triggers a 500 (or even a 200 response), and commit this contract to a repo even if the resulting contract test fails. We wish to keep the example so that the feedback is not lost, but cannot afford to break the build as the developer needs time to fix the problem.

**How**:

- The Row class now accepts a name.
- When generating a test from a row, Specmatic now checks if it is a WIP row, if so sets the ignoreFailures flag on the test.
- Specmatic does not validate an example if it is for a 400 or 422 response. This works only for examples inside a yaml file.

**Checklist**:

- [/] Tests
- [/] Sonar Quality Gate

**Issue ID**:
Closes: #123

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
@@ -249,7 +249,8 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
                                             ObjectMapper().readValue(valueString, Map::class.java).values.first()
                                                 .toString()
                                         } else valueString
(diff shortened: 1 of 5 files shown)


harikrishnan83 (approved):
Looks good to me.

Thanks.

PR #626, opened 2023-03-17
Avoid combinatorial explosion during test generation

### LINKED ISSUE #625: Combinatorial explosion during test generation
**Description**
Combinatorial explosion during test generation
- Wait-time and used memory become unacceptable even in basic cases, since the current implementation produces the cartesian product of each request property's possible values.
- For example, 16K combinations are produced if there is a request object with seven properties with 4 possible values each. Another real-world example endpoint has produced over 35 million combinations.

**Steps to reproduce**
1. Run [this unit test](https://github.com/znsio/specmatic/pull/626/files#diff-69e38caff576912f4879df48828cb2543a3408ef82fdb0bf209dcaba2b9563e5R116) from PR #626 without any of the other changes.
2. Notice extremely long durations and, likely, OutOfMemoryError
3. Apply the rest of the changes from PR #626 and run the test again
4. Notice the test pass quickly

**Expected behavior**
Produce as many test combinations as possible up to a configured max, while also ensuring every candidate value is tested in at least one test combination.

**System Information:**
 - Specmatic version: [e.g. 0.63.1]

westse:
**What**: Avoids combinatorial explosion during test generation

**Why**: Fixes #625 . Without this wait-time and used memory become unacceptable even in basic cases, since the previous implementation produces the cartesian product of each request property's possible values.

**How**: Introduces CombinationSpec to support producing as many test combinations as possible up to a configured max, while also ensuring every candidate value is tested in at least one test combination.

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #625

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Flags.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Flags.kt
@@ -5,6 +5,7 @@ import org.apache.commons.lang3.BooleanUtils
 object Flags {
     private const val customResponseName = "CUSTOM_RESPONSE"
     const val negativeTestingFlag = "SPECMATIC_GENERATIVE_TESTS"
+    const val maxTestRequestCombinationsFlag = "MAX_TEST_REQUEST_COMBINATIONS"
 
     private fun flagValue(flagName: String): String? {
         return System.getenv(flagName) ?: System.getProperty(flagName)
@@ -16,5 +17,9 @@ object Flags {
 
     fun generativeTestingEnabled(): Boolean {
         return BooleanUtils.toBoolean(flagValue(negativeTestingFlag) ?: "false")
+    }
+
+    fun maxTestRequestCombinations(): Int {
+        return (flagValue(maxTestRequestCombinationsFlag) ?: "64").toInt()
     }
 }
--- a/core/src/main/kotlin/in/specmatic/core/Flags.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Flags.kt
@@ -20,6 +20,6 @@ object Flags {
     }
 
     fun maxTestRequestCombinations(): Int {
-        return (flagValue(maxTestRequestCombinationsFlag) ?: "64").toInt()
+        return flagValue(maxTestRequestCombinationsFlag)?.toInt() ?: Int.MAX_VALUE
     }
 }
--- a/core/src/main/kotlin/in/specmatic/core/pattern/CombinationSpec.kt
+++ b/core/src/main/kotlin/in/specmatic/core/pattern/CombinationSpec.kt
@@ -1,0 +1,83 @@
+package `in`.specmatic.core.pattern
+
+import kotlin.math.min
+
+/**
+ * Provides utility access to all combinations of multiple sets of candidate values. Represented as the cartesian
+ * product of all sets where each combination is represented as a numerical index from 0 to (MAX_COMBOS - 1).
+ *
+ * Supports sequential iteration over all combinations, which is used to test as many combinations as possible.
+ *
+ * Supports translating a combination of candidate values to an index, which is used to first test priority combinations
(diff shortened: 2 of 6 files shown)


westse:
Can this be merged? Any update is appreciated!

harikrishnan83:
@westse we will take a look at this shortly. Thanks for the nudge.

joelrosario:
Hi @westse , our apologies for the wait, and we appreciate your patience.

We have users that wish to generate all the combinations and make all the tests pass, or in some cases do so as part of exploratory investigations.

To accommodate this, can you make the max count opt-in. In other words:
- users that wish to run all the tests simply need not specify set a value for `MAX_TEST_REQUEST_COMBINATIONS`,
- while users who prefer not to run all the tests may provide a value for `MAX_TEST_REQUEST_COMBINATIONS`.

joelrosario (changes requested):
Hi @westse , our apologies for the wait, and we appreciate your patience.

We have users that wish to generate all the combinations and make all the tests pass, or in some cases do so as part of exploratory investigations.

To accommodate this, can you make the max count opt-in. In other words:
- users that wish to run all the tests simply need not specify set a value for `MAX_TEST_REQUEST_COMBINATIONS`,
- while users who prefer not to run all the tests may provide a value for `MAX_TEST_REQUEST_COMBINATIONS`.

westse:
> We have users that wish to generate all the combinations and make all the tests pass, or in some cases do so as part of exploratory investigations.
> 
> To accommodate this, can you make the max count opt-in. In other words:
> 
> * users that wish to run all the tests simply need not specify set a value for `MAX_TEST_REQUEST_COMBINATIONS`,
> * while users who prefer not to run all the tests may provide a value for `MAX_TEST_REQUEST_COMBINATIONS`.

@joelrosario Thanks for the response. It is a good point that different users will have different tolerances (I also want to run as many combinations as I can! :grinning: ). Rather than defaulting to opt-out, would it instead make sense to default `MAX_TEST_REQUEST_COMBINATIONS` significantly higher? It seems we should be able to pick a value higher than needed by any user.

For example, lets say we default to `Int.MAX_VALUE` (`2,147,483,647`) and that we actually have at least that many combinations to test with (e.g. 10 params with an average of 9 possible values each, or `9^10=3,486,784,401`). Assuming specmatic takes 100ms to test each combination (very optimistic) then that will take 6.8 years to complete, assuming my math is correct (`2,147,483,647 combos / 10 combos per second / 60 seconds / 60 minutes / 24 hours / 365 days = 6.80962597349 years`). If we agree that is likely unacceptable to any user, then using `Int.MAX_VALUE` is effectively opting out, and could probably even use a much lower value.

My own experience in case it is helpful: I picked 64 since even that seemingly low value resulted in about 3x the time I deemed acceptable for testing a single API endpoint, and we have many such endpoint to test! So I am already explicitly setting `MAX_TEST_REQUEST_COMBINATIONS` lower (to 24) and not taking the default of 64.

joelrosario:
> For example, lets say we default to `Int.MAX_VALUE` (`2,147,483,647`) and that we actually have at least that many combinations to test with (e.g. 10 params with an average of 9 possible values each, or `9^10=3,486,784,401`). Assuming specmatic takes 100ms to test each combination (very optimistic) then that will take 6.8 years to complete, assuming my math is correct (`2,147,483,647 combos / 10 combos per second / 60 seconds / 60 minutes / 24 hours / 365 days = 6.80962597349 years`). If we agree that is likely unacceptable to any user, then using `Int.MAX_VALUE` is effectively opting out, and could probably even use a much lower value.

Thanks @westse, that's a good idea, setting a max value of Int.MAX_VALUE works just as well for us.

westse:
@joelrosario Ok, that change is done and pushed - thanks.

joelrosario:
Looks good, thanks @westse !

PR #655, opened 2023-05-19
Loading contract paths bug fix

jaydeepk:
Fixed bug wherein an error is throw while loading contract paths, when the repos base directory exists, but the actual contract repo directory does not exist.


joelrosario (approved):
Reviewed over a pairing session, looks good to me

PR #662, opened 2023-06-03
Prevent sending a header by marking it as omit in open api…

harikrishnan83:
**What**:

Prevent sending header in the request by marking it as "(OMIT)" in example value

**Why**:

Currently when there is no example in OpenAPI, the header value is auto generated and sent as part of contract test. Certain test cases require a header to be completely not sent at all.

Example: A test case to verify if the server returns a 400 BAD REQUEST if a mandatory header such as traceId is missing will require us to deliberately not send the header as part of the generated test.

**How**:

Leveraging "(OMIT)" keyword which when provided as the example value for a header will prevent that header from being sent.

**Checklist**:

- [ ] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation - Pending
- [x] Tests
- [x] Sonar Quality Gate


joelrosario (approved):
LGTM

PR #664, opened 2023-06-07
Add Bearer prefix to Authentication header when using bearer security scheme

androa:
**What**: Add `Bearer` prefix to the `Authorization` header when API is using `Bearer` based authentication.

**Why**: When setting up an authentication scheme using `bearer` as `scheme` the requests made when contract testing should contain a `Bearer` prefix in the generated `Authentication` header.

Many frameworks, `ktor` included, requires the `Bearer` prefix to be there to correctly delegate to the specified auth handler. Without it `ktor` automatically rejects the request with a `401 Unauthorized` when the `Authorization` header simply is `Authorization: WFPSL`.

**How**: Add a `Bearer` prefix to the `Authorization` header.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/BearerSecurityScheme.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/BearerSecurityScheme.kt
@@ -9,7 +9,9 @@ import org.apache.http.HttpHeaders.AUTHORIZATION
 
 class BearerSecurityScheme : OpenAPISecurityScheme {
     override fun matches(httpRequest: HttpRequest): Boolean {
-        return httpRequest.headers.containsKey(AUTHORIZATION)
+        return httpRequest.headers.let {
+            it.containsKey(AUTHORIZATION) && it[AUTHORIZATION]!!.lowercase().startsWith("bearer")
+        }
     }
 
     override fun removeParam(httpRequest: HttpRequest): HttpRequest {
--- a/core/src/main/kotlin/in/specmatic/conversions/BearerSecurityScheme.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/BearerSecurityScheme.kt
@@ -3,15 +3,20 @@ package `in`.specmatic.conversions
 import `in`.specmatic.core.HttpRequest
 import `in`.specmatic.core.HttpRequestPattern
 import `in`.specmatic.core.Resolver
+import `in`.specmatic.core.Result
 import `in`.specmatic.core.pattern.Row
 import `in`.specmatic.core.pattern.StringPattern
 import org.apache.http.HttpHeaders.AUTHORIZATION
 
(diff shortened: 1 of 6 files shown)


joelrosario:
Hi @androa,

Firstly, thanks a lot for contributing this PR!

While this takes care of contract testing, consumers would need similar feedback, and so the same constraint needs be implemented in Specmatic's stub.

This means that when validating an expectation with the `Authorisation` header, where the APIs security scheme is `bearer`, Specmatic should show an error against the `Authorisation` header if it's value does not start with `Bearer `.

We'll keep you posted on this.

core/src/main/kotlin/in/specmatic/conversions/BearerSecurityScheme.kt
```diff
@@ -9,7 +9,9 @@ import org.apache.http.HttpHeaders.AUTHORIZATION
 
 class BearerSecurityScheme : OpenAPISecurityScheme {
     override fun matches(httpRequest: HttpRequest): Boolean {
-        return httpRequest.headers.containsKey(AUTHORIZATION)
+        return httpRequest.headers.let {
+            it.containsKey(AUTHORIZATION) && it[AUTHORIZATION]!!.lowercase().startsWith("bearer")
```
androa: @joelrosario Would that be something like this? It is rather awkward to not get a better error than the generic `No auth params were found in the request`, but I'm not quite sure the best way to change the signature of `matches()` in a way it can create different failure responses.

joelrosario: @androa yes, and thanks, that change looks good!

But you're also right about the generic message not being too helpful.

We'd use Specmatic's Result class to provide more detail.

This means, we'd need to change the return type of the `matches` method from Boolean to Result.

I believe there should not be much impact as this method is not referred to in many places, which include:
1. the other classes implementing OpenAPISecurityScheme
2. `HttpRequestPattern.matchSecurityScheme`... we'd need to modify a line of code in that method to match the new signature.

androa: Made an attempt of refactoring it into returning the Result class now.

androa: I'm having some issues with running the test suite locally so the failing test in OpenApiKtTest slipped past me. Not quite sure what causes it though.

androa:
Any chance you could do a review of this change?

harikrishnan83:
@androa apologies for the delay, we will look into this at the earliest. cc @jaydeepk @joelrosario

harikrishnan83:
bump

joelrosario (approved):
Looks good to me! Thanks @androa for your patience and for the PR, really appreciate it.

PR #594, opened 2023-01-23
Support oneOf w/ discriminator

westse:
**What**: Adds support for discriminator per [OAS spec](https://spec.openapis.org/oas/latest.html#discriminator-object). NOTE unit test depends on #585 having been fixed.

**Why**: To fix issue #593 and add discriminator support per OAS spec.

**How**: Enhanced existing allOf support

**Checklist**:

- [ ] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation
- [X] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #593

NOTE unit test depends on #585 having been fixed

PR #667, opened 2023-06-10
Fix to discriminator implementation

joelrosario:
**What**:

The discriminator currently is read from the object itself. However, according to the OpenAPI documentation, the discriminator property is in the same schema object that contains oneOf or anyOf.

Reference: https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
@@ -854,6 +854,8 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
     ) = schema.properties.orEmpty().map { (propertyName, propertyType) ->
         if (discriminatorPropertyName == propertyName)
             propertyName to ExactValuePattern(StringValue(patternName))
+        else if(schema.discriminator?.propertyName == propertyName)
+            propertyName to ExactValuePattern(StringValue(patternName))
         else {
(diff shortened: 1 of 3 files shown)


westse (commented):
@joelrosario I saw your comment on issue #594 , and I agree. In fact I believe I have been confused in the past about this, but figured I was either incorrect or the the impact wasn't critical yet since the discriminator is used as an optimization for consumers.

In this PR it appears you have left in the previous support for in-schema discriminator. Is that for backward compatibility? If so, that makes sense, but I wonder if it instead should be dropped by default, but then allow opt-in via explicit property. That way in the future any new and unaffected consumers won't inadvertently depend on the out-of-spec behavior.

joelrosario:
Thanks @westse for checking out the PR!

> In this PR it appears you have left in the previous support for in-schema discriminator. Is that for backward compatibility? If so, that makes sense, but I wonder if it instead should be dropped by default, but then allow opt-in via explicit property. That way in the future any new and unaffected consumers won't inadvertently depend on the out-of-spec behavior.

I left it in until we'd had a chance to discuss this with you. I'm thinking about either turning it off by default or just removing it. I'll discuss this with @harikrishnan83 and keep you posted.

harikrishnan83:
@westse apologies for the long radio silence on this issue. We are working on an alternate fix for the discriminator issue. So closing this PR for now. cc @joelrosario

PR #687, opened 2023-07-21
Fixed url paths not covered in api coverage report

vikram-rao:
**What:** Url path with multiple path parameters is not parsed correctly due to regex pattern matching error during parsing (requires non-greedy matching)

**Why:** Api Coverage is incorrect due to wrong path string manipulation. All apis with multiple path params are marked as not covered

**How:** Simple modification of the regex used for path conversion to be non greedy using `?` operator

**Checklist**: 

- [ ] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation : N/A
- [x] Tests
- [ ] Sonar Quality Gate : N/A

**Issue ID**:
N/A

### DIFF
--- a/core/src/main/kotlin/in/specmatic/test/ScenarioTest.kt
+++ b/core/src/main/kotlin/in/specmatic/test/ScenarioTest.kt
@@ -6,7 +6,7 @@ import `in`.specmatic.core.executeTest
 
 class ScenarioTest(val scenario: Scenario, private val generativeTestingEnabled: Boolean = false) : ContractTest {
     override fun testResultRecord(result: Result): TestResultRecord {
-        return TestResultRecord(scenario.path.replace(Regex("""\((.*):.*\)"""), "{$1}"), scenario.method, scenario.status, result.testResult())
+        return TestResultRecord(scenario.path.replace(Regex("""\((.*?):.*?\)"""), "{$1}"), scenario.method, scenario.status, result.testResult())
     }
 
     override fun generateTestScenarios(
--- a/core/src/main/kotlin/in/specmatic/test/ScenarioTest.kt
+++ b/core/src/main/kotlin/in/specmatic/test/ScenarioTest.kt
@@ -1,12 +1,13 @@
 package `in`.specmatic.test
 
+import `in`.specmatic.conversions.convertPathParameterStyle
 import `in`.specmatic.core.Result
 import `in`.specmatic.core.Scenario
 import `in`.specmatic.core.executeTest
 
 class ScenarioTest(val scenario: Scenario, private val generativeTestingEnabled: Boolean = false) : ContractTest {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/in/specmatic/test/ScenarioTestGenerationFailure.kt
```diff
@@ -6,7 +6,7 @@ import `in`.specmatic.core.Result
 
 class ScenarioTestGenerationFailure(val scenario: Scenario, val e: Throwable) : ContractTest {
     override fun testResultRecord(result: Result): TestResultRecord {
-        return TestResultRecord(scenario.path.replace(Regex("""\((.*):.*\)"""), "{$1}"), scenario.method, scenario.httpResponsePattern.status, result.testResult())
+        return TestResultRecord(scenario.path.replace(Regex("""\((.*?):.*?\)"""), "{$1}"), scenario.method, scenario.httpResponsePattern.status, result.testResult())
```
joelrosario: Can you add a unit test for this method

core/src/main/kotlin/in/specmatic/test/ScenarioTest.kt
```diff
@@ -6,7 +6,7 @@ import `in`.specmatic.core.executeTest
 
 class ScenarioTest(val scenario: Scenario, private val generativeTestingEnabled: Boolean = false) : ContractTest {
     override fun testResultRecord(result: Result): TestResultRecord {
-        return TestResultRecord(scenario.path.replace(Regex("""\((.*):.*\)"""), "{$1}"), scenario.method, scenario.status, result.testResult())
+        return TestResultRecord(scenario.path.replace(Regex("""\((.*?):.*?\)"""), "{$1}"), scenario.method, scenario.status, result.testResult())
```
joelrosario: Can you add a unit test for this method

junit5-support/src/main/kotlin/in/specmatic/test/SpecmaticJUnitSupport.kt
```diff
@@ -110,7 +110,7 @@ class TestReport(private val testReportRecords: MutableList<TestResultRecord> =
         logger.newLine()
 
         val recordsWithFixedURLs = testReportRecords.map {
-            it.copy(path = it.path.replace(Regex("""\((.*):.*\)"""), "{$1}"))
+            it.copy(path = it.path.replace(Regex("""\((.*?):.*?\)"""), "{$1}"))
```
joelrosario: Can you extract this functionality to a method and add a unit test.

The functionality of doing the regex replace across all 3 seems the same.. if so, can you extract the regex replace itself into a common method.

joelrosario (changes requested):

joelrosario (approved):
Looks good, thanks @vikram-rao!

PR #700, opened 2023-08-10
Allow command install to be set in a specific directory even in a Git directory + handle the branch in install command

Snorkell:
**What**: I added the feature to allow the user to specify where the install command should place the files from the OpenApi repository, allowing it to be placed inside another directory which is also a git directory (nested repository)
I added the feature to change the branch by using the one specified in the `specmatic.json` whe using the command install

**Why**: Install by default was installing in the home directory which can be an issue while working with docker container and k8s pods, or even with a CI tool a bit too old like bamboo

**How**: 
- I added an option to the command `install` which is called `--targetDirectory` wich by default is home if not specified
- I used the clone command from the `GitRepo.kt` instead of the `sourceGit.clone(....)` to handle the branch specified in the `specmatic.json`
- I check that if the directory is a git directory but the remote is not the same and the sub direcotries are empty or the directory is not a git directory then clone from the repo

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: /

### DIFF
--- a/application/src/test/kotlin/application/FakeGit.kt
+++ b/application/src/test/kotlin/application/FakeGit.kt
@@ -99,4 +99,7 @@ abstract class FakeGit: GitCommand {
     override fun revisionsBehindCount(): Int {
         TODO("Not yet implemented")
     }
+    override fun getRemoteUrl(name: String): String {
+        TODO("Not yet implemented")
+    }
 }
--- a/core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
+++ b/core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
@@ -117,7 +117,7 @@ data class GitRepo(
             if (!sourceDir.exists())
                 sourceDir.mkdirs()
 
-            if (!sourceGit.workingDirectoryIsGitRepo() || (sourceGit.workingDirectoryIsGitRepo() && sourceGit.getRemoteUrl() != this.gitRepositoryURL && sourceDir.listFiles()?.isEmpty()!!)) {
+            if (!sourceGit.workingDirectoryIsGitRepo() || isEmptyNestedGitDirectory(sourceGit, sourceDir)) {
                 println("Found it, not a git dir, recreating...")
                 sourceDir.deleteRecursively()
                 sourceDir.mkdirs()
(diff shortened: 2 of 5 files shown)


core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
```diff
@@ -117,12 +117,12 @@ data class GitRepo(
             if (!sourceDir.exists())
                 sourceDir.mkdirs()
 
-            if (!sourceGit.workingDirectoryIsGitRepo()) {
+            if (!sourceGit.workingDirectoryIsGitRepo() || (sourceGit.workingDirectoryIsGitRepo() && sourceGit.getRemoteUrl() != this.gitRepositoryURL && sourceDir.listFiles()?.isEmpty()!!)) {
```
harikrishnan83: Can we extract `(sourceGit.workingDirectoryIsGitRepo() && sourceGit.getRemoteUrl() != this.gitRepositoryURL && sourceDir.listFiles()?.isEmpty()!!)` to a method that is named such that it conveys the purpose of these conditional checks? Just to aid readability.

harikrishnan83:
@Snorkell Thanks for your PR. Really appreciate the effort. We will review you PR shortly.

harikrishnan83 (changes requested):

core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
Snorkell: Thanks for the feedback, i'm fixing this right now

application/src/test/kotlin/application/FakeGit.kt
```diff
@@ -99,4 +99,7 @@ abstract class FakeGit: GitCommand {
     override fun revisionsBehindCount(): Int {
         TODO("Not yet implemented")
     }
+    override fun getRemoteUrl(name: String): String {
```
joelrosario, line 102: Can you rename this to remoteName.

The git command param is called "name", however that is in the context of the remote command so the meaning of name is clearer. This is in the context of the GitCommand so it may not be that obvious.

joelrosario: Fixed it myself while doing the pull request

joelrosario (approved):
Looks good to me.

harikrishnan83 (approved):

joelrosario:
Thanks @Snorkell, really appreciate the PR!

PR #712, opened 2023-09-01
fix: correctly filter schema scenarios already defined in external spec

### LINKED ISSUE #711: External test case filters out unrelated spec scenarios
**Description**
External spec filters out unrelated scenarios parsed from API schema

**Steps to reproduce**
Create a schema with multiple `GET` endpoints that return 200 response e.g. 
Create a spec file with single scenario for 1 of the endpoints of the schema, that asserts a returned status using `Then status 200`.   
When the test runs, all scenarios from the API schema that have the same response code will be ignored and no test case will be generated for them (with the exception of the one that I actually describe in the gherkin spec).
This is most probably due to filtering in: https://github.com/znsio/specmatic/blob/b66b03c9e360f0a7f38668f6ed03011af97ea1b5/core/src/main/kotlin/in/specmatic/core/Feature.kt#L1657-L1661
where **all** `scenarioInfosBelongingToIncludedSpecifications`  are filtered even if there is no scenario in my spec that actually describes that endpoint. 

**Expected behavior**
The logic should not filter out unrelated endpoints - only the one with matching request pattern.

matus-m:
**What**:

This should fix the overly aggressive filter in Feature scenarios, that removes any Schema scenario that has the same return status value as a scenario defined in external spec file. 

**Why**:

See #711 

**How**:

Filter now consider both the status and request pattern match

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate 

**Issue ID**:
Closes: #711

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Feature.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Feature.kt
@@ -1655,7 +1655,7 @@ fun scenarioInfos(
     }
 
     return specmaticScenarioInfos.plus(scenarioInfosBelongingToIncludedSpecifications.filter { scenarioInfo ->
-        !specmaticScenarioInfos.any {
+        specmaticScenarioInfos.none {
             it.httpResponsePattern.status == scenarioInfo.httpResponsePattern.status
                     && it.httpRequestPattern.matchesSignature(scenarioInfo.httpRequestPattern)
         }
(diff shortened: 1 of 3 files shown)


joelrosario (approved):
Thanks @yntelectual for the PR. Appreciate the fix!

Made a small update to the test. Looks good to me.

PR #716, opened 2023-09-04
Add String format: byte pattern support

aeoncl:
**What**: Added support for OpenAPI string byte pattern

**Why**: Sometimes in life, you encounter an openAPI spec with a string byte property:

```yaml
myString:
  type: string
  format: byte
```

**How**:

We parsed the OpenAPI ByteArrayPattern to Specmatic's BinaryPattern

**Checklist**:

- [X] Tests
- [ ] Sonar Quality Gate N/A

**Issue ID**:
Closes: /


core/src/test/kotlin/in/specmatic/conversions/OpenApiSpecificationTest.kt
```diff
@@ -6467,6 +6468,63 @@ paths:
...
+            val result =
+                    this.scenarios.first().matchesMock(
+                            HttpRequest(
+                                    "POST",
+                                    "/file",
+                                    multiPartFormData = listOf(MultiPartFileValue("filesPart", "test.pdf", "application/pdf", "UTF-8"))
+                            ), HttpResponse.OK("{\"filename\": \"ThIsi5ByT3sD4tA\"}")
+                    )
+
+            println(result.reportString())
```
harikrishnan83: Can we please remove the print statement?

harikrishnan83:
@aeoncl Thanks for your PR. We will take a look.

joelrosario:
Hi @aeoncl, thanks a lot for contributing this PR!

I had a close look at the documentation for strings with the format byte. It can be found on [this page](https://swagger.io/docs/specification/data-models/data-types/). Please have a look.

According to the documentation, byte and binary are different types. Specmatic would need to have new Pattern class (ByteStringPattern perhaps?) to match and generate base64 strings.

Please have a look at the documentation and let us know what you think.

core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -627,6 +627,7 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
                     throw UnsupportedOperationException("Unsupported composed schema: $schema")
                 }
             }
+            is ByteArraySchema -> BinaryPattern()
```
joelrosario: We'd need to implement a new Pattern class for ByteArraySchema here. See my comment at https://github.com/znsio/specmatic/pull/716#issuecomment-1722482596.

joelrosario (changes requested):

core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
aeoncl: Yes, i came to the same conclusion while working with this quick fix: it needs a new Pattern with it's own matching logic & generation.

It would be cool to be able to use externalValue for stubbing with this Pattern Class (taking the file bytes and encoding them to base64)
Currently if i understand correctly: 

When dealing with externalValue , the parser adds _filename to the key that corresponds to the file path, then somewhere in the response generation, the file content is added to the response without looking at the pattern type.

Maybe we'll need to adapt this behaviour to support different externalValue treatment based on the PatternType?

aeoncl: After looking further, i realized that externalValue is not supported for the responses yet. i did not find anything that loads file content in the response. This could be useful if we want to send a string of bytes or just return an application octet stream, i'm gonna create an issue for this

joelrosario: @aeoncl that's right, Specmatic does not yet support for `externalValue`.

We have been using examples in the specification file, as we find that keeping the example inline tends to make it easier to understand the request format.

However I also see what you mean and I can see that `externalValue` ([documented here](https://swagger.io/docs/specification/adding-examples/)) could be useful if you need to use a base64 encoded file as an example.

joelrosario (changes requested):
Hi @aeoncl, thanks, looks good!

I've added a test and slightly modified a line of code. I've given reasons for this in the respective commit messages. Hope this is fine.

Could you add 2 integration tests:
1. a test with a specification containing a byte string with an base64 example running as test (see tests in the `AuthorizationCodeFlowTest` class for ideas on how to do this)
2. a test with a specification containing a byte string with an expectation running as stub (see the test named `should create stub from gherkin that includes OpenAPI spec` for an example of how you could do this)

aeoncl:
Nice catch ! Okay i'll look into adding those

joelrosario (approved):
Thanks for the PR @aeoncl and for your patience. I've added some tests, and made a few minor changes to Base64StringPattern.

Merging this PR.

PR #725, opened 2023-09-12
OAuth support

### LINKED ISSUE #611: Oauth2 Authorization
**Is your feature request related to a problem? Please describe.**
Specmatic only supports bearer and api key authentication (header, query) security schemes at the moment

**Describe the solution you'd like**
Will specmatic support Oath2 security https://swagger.io/docs/specification/authentication/oauth2/ ?

How can I still use specmatic with specs that have oath2 authorization?
Can i ignore the error message and uses bearer tokens myself by fetching a token in beforeAll without changing the actual schema?

Thank you for your help!

jaydeepk:
**What**:
Implemented support for OAuth2 Security Scheme.
Implemented a 'security' section in specmatic.json to define tokens for OAuth, Bearer, and API Key security schemes.
Specmatic will then use the specified tokens while executing contract tests.

- [x] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #611


harikrishnan83 (approved):
Looks good to me

joelrosario (approved):
Looks good to me, thanks @jaydeepk !

androa:
Tested this now. It seems to work as expected when I set an actual environment variable. But I'm trying to run a contract test programmatically with jUnit, and need to calculate the token runtime, so I thought I could set it as a property.

As far as I can see on the code, that should be supported as well. But it does not work. I suspect it might be the caching mechanism implemented in `EnvironmentAndPropertiesConfiguration()` and that the instantiation of `SpecmaticJUnitSupport` happens before my contract test file.

Any ideas on cause or how I can debug this better?

androa:
I'm trying to create a token in conjunction with `SpecmaticJUnitSupport`. I want to be able to create a token at runtime, before starting the tests.

But https://github.com/znsio/specmatic/blob/4caabcfcfcec45cb7717b21c7b8ba99979239b60/core/src/main/kotlin/in/specmatic/conversions/SecurityToken.kt#L28 checks only environment. Is there any reasons for that, or could it simply call `getCachedSetting()` instead?

I believe that would work in my case. Created an PR for this https://github.com/znsio/specmatic/pull/1055

PR #737, opened 2023-09-20
Stub usage report implementation

jaydeepk:
**What**:

Implemented the generation of a 'Stub Usage Report' json file.
The report will be automatically generated once the stub process is stopped.
The report will list all the endpoints for which the stub returned a successful response (in a format similar to the API Coverage Report).

- [x] Tests
- [ ] Sonar Quality Gate


core/src/main/kotlin/in/specmatic/core/HttpResponse.kt
```diff
@@ -35,6 +35,8 @@ data class HttpResponse(
     fun specmaticResultHeaderValue(): String =
         this.headers.getOrDefault(SPECMATIC_RESULT_HEADER, "success")
 
+    fun isSuccessful(): Boolean = specmaticResultHeaderValue() == "success"
```
harikrishnan83: @joelrosario is there another way to check if a response is a success response? also just curious if the result header value is a dependable way of checking if a response is a success response. cc @jaydeepk

core/src/main/kotlin/in/specmatic/stub/HttpStub.kt
```diff
@@ -427,6 +459,40 @@ class HttpStub(
 
     init {
         server.start()
+        extractALlEndpoints()
```
joelrosario: I think this method can be called when _allEndpoints is initialised, and then _allEndpoints can become immutable

core/src/main/kotlin/in/specmatic/core/HttpResponse.kt
```diff
@@ -35,6 +35,8 @@ data class HttpResponse(
     fun specmaticResultHeaderValue(): String =
         this.headers.getOrDefault(SPECMATIC_RESULT_HEADER, "success")
```
joelrosario: Should we not default to "failure"? If the header does not exist it means failure.

core/src/main/kotlin/in/specmatic/core/HttpResponse.kt
joelrosario: I think there's a way we can use the type system to signal success or failure.

The code is structured to return a stubbed response. However it's possible to wrap the HttpStubResponse object in another object that contains the behaviour for logging (or not logging as the case may be).

The most expeditious way would be to create a wrapper function around HttpStub/getHttpResponse (or alter it's signature), then introduce the wrapper object in getHttpResponse at line 624, and in fakeHttpResponse at lines 756 and 758.

cc @jaydeepk

joelrosario (changes requested):
I've left some comments.

core/src/main/kotlin/in/specmatic/stub/HttpStub.kt
```diff
@@ -65,9 +67,21 @@ class HttpStub(
...
 
+    companion object {
+        const val JSON_REPORT_PATH = "./build/reports/specmatic"
+        const val JSON_REPORT_FILE_NAME = "stub_usage_report.json"
+    }
+
     private val threadSafeHttpStubs = ThreadSafeListOfStubs(_httpStubs.filter { it.stubToken == null }.toMutableList())
     private val threadSafeHttpStubQueue = ThreadSafeListOfStubs(_httpStubs.filter { it.stubToken != null }.reversed().toMutableList())
 
+    private val _logs: MutableList<StubEndpoint> = mutableListOf()
```
joelrosario: This data structure needs to be made thread-safe

joelrosario (changes requested):

core/src/main/kotlin/in/specmatic/core/HttpResponse.kt
joelrosario: This is also used elsewhere in the code in the context of which this behaviour is correct. Closing this.

core/src/main/kotlin/in/specmatic/stub/HttpStub.kt
jaydeepk: Done

core/src/main/kotlin/in/specmatic/stub/HttpStub.kt
```diff
@@ -65,9 +67,21 @@ class HttpStub(
...
 
+    companion object {
+        const val JSON_REPORT_PATH = "./build/reports/specmatic"
+        const val JSON_REPORT_FILE_NAME = "stub_usage_report.json"
+    }
+
     private val threadSafeHttpStubs = ThreadSafeListOfStubs(_httpStubs.filter { it.stubToken == null }.toMutableList())
     private val threadSafeHttpStubQueue = ThreadSafeListOfStubs(_httpStubs.filter { it.stubToken != null }.reversed().toMutableList())
 
+    private val _logs: MutableList<StubEndpoint> = Collections.synchronizedList(ArrayList())
```
joelrosario, line 78: There should be a test for this. Have say 10 threads running simultaneously making requests, and we should have 10 logs in this list.

joelrosario (changes requested):

core/src/main/kotlin/in/specmatic/core/HttpResponse.kt
harikrishnan83: @jaydeepk @joelrosario okay, if this is dependable way of checking if the response is a success then looks good.

core/src/main/kotlin/in/specmatic/stub/HttpStub.kt
jaydeepk: Sure, I have added the test.

joelrosario (approved):
Looks good to me, thanks @jaydeepk

PR #739, opened 2023-09-21
Don't delete .specmatic folder

vikram-rao:
Don't delete .specmatic folder to be consistent in this behaviour in all places

**What**: .specmatic folder is removed after the command executes in some situations and not in others. We want to keep this consistent. It was decided to retain it once created.

**Why**: Consistency and also ideally some one may want to refer to it after command executes too

**How**: Across all places where workspace directory is deleted (HttpStub, JunitSupport and the delete method itself in WorkingDirectory)

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate

**Issue ID**:


core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
```diff
@@ -94,7 +96,7 @@ data class GitRepo(
     }
 
     private fun isClean(contractsRepoDir: File): Boolean {
-        val sourceGit = getSystemGit(contractsRepoDir.path)
+        val sourceGit = getSystemGitWithAuth(contractsRepoDir.path)
```
joelrosario: use getSystemGit where auth is not actually required.

core/src/main/kotlin/in/specmatic/core/utilities/ContractSource.kt
```diff
@@ -103,18 +105,51 @@ data class GitRepo(
...
-
         when (branchName) {
             null -> logger.log("No branch specified, using default branch")
             else -> checkout(repositoryDirectory, branchName)
         }
+        ensureThatSpecmaticFolderIsIgnored()
         return repositoryDirectory
     }
 
+    private fun ensureThatSpecmaticFolderIsIgnored() {
```
joelrosario, line 127: can be simplified as discussed

joelrosario (approved):
Looks good to me

PR #753, opened 2023-09-28
removing explicit kotlin compile jvm target to 1.8

harikrishnan83:
**What**:

Removing explicit kotlin compile jvm target to 1.8

**Why**:

CodeQL and Sonar builds are failing with ```Caused by: org.gradle.api.GradleException: 'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version```

**How**:

The JDK on the build machines are set to 1.8 and release also ensure 1.8 JDK, so there is not need to explicitly set target JVM to 1.8

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -100,12 +100,6 @@ jacocoTestReport {
 
 group = 'in.specmatic'
 
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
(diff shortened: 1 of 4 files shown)


joelrosario (approved):
Looks good to me.

PR #781, opened 2023-10-08
Resolve refs to parameters and request / response body examples

joelrosario:
**What**:

1. OpenAPI allows the definitions of headers, query parameters, path parameters, request and response bodies to `$ref` to the components section instead of containing the definition inline.

For example, a request body definition could `$ref` to a component in the `requestBodies` section like this:

```yaml
paths:
  /person:
    post:
      summary: "Get person by id"
      parameters: []
      requestBody:
        ${"$"}ref: '#/components/requestBodies/PersonRequest'
components:
  requestBodies:
    PersonRequest:
      content:
        application/json:
          schema:
            required:
            - "id"
            properties:
              id:
                type: "string"
          examples:
            200_OK:
              value:
                id: "abc123"
```

Specmatic can now understand these kinds of `$ref`s for the definitions of parameters (header, query and path) and request and response bodies.

Support for form fields and multi-part requests will be implemented in a future PR.

2. Additionally, nested allOfs are now supported. Previously, if an allOf contains another allOf, the nested allOf would be completely ignored.

**Why**:

We have encountered several specifications recently such as those on the [TMForum ODA OpenAPI Table page](https://www.tmforum.org/oda/open-apis/table) that are defined this way.

**How**:

Specmatic now looks for `$ref`s in the previously mentioned types of components and resolves them where found.

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/in/specmatic/conversions/OpenApiSpecification.kt
@@ -621,6 +621,18 @@ class OpenApiSpecification(private val openApiFile: String, val openApi: OpenAPI
     fun toSpecmaticPattern(mediaType: MediaType, jsonInFormData: Boolean = false): Pattern =
         toSpecmaticPattern(mediaType.schema, emptyList(), jsonInFormData = jsonInFormData)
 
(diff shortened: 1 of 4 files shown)


harikrishnan83 (approved):
Looks good to me

PR #782, opened 2023-10-08
Improve test name filter

joelrosario:
**What**:

This PR does the following
1. filters the tests before the test generation (see more on this in the how section)
2. logs descriptions of tests that were selected, if any tests were filtered out
3. fixes bug in the filteration logic

**Why**:

We saw some cases where we needed to leave the spec as pristine as possible and run tests for only 1 or 2 specific APIs.

**How**:

Tests are generated in two steps.
1. Load the operations from the specification file
2. Generate tests from the examples in each of the operations

The filtration was done after step 2, which means that it was done after all the permutations were derived.

With this PRs, the filtration logic runs after step 1 as well, before any of the permutations.

It still runs after step 2, in case any of the filter terms include test names. Test names are not available yet after step 1 given that the examples have not been taken into account.

**Checklist**:

- [x] Tests

### DIFF
--- a/application/src/test/kotlin/in/specmatic/test/SpecmaticJUnitSupportKtTest.kt
+++ b/application/src/test/kotlin/in/specmatic/test/SpecmaticJUnitSupportKtTest.kt
@@ -55,14 +55,14 @@ paths:
 
     @Test
     fun `should select tests containing the value of filterName in testDescription`() {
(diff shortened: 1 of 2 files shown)


harikrishnan83 (approved):
Looks good to me

PR #793, opened 2023-10-17
Fixes to oneOf generation and error messages

joelrosario:
**What**:

* There a bug fix to generation of oneOf. It was breaking when generative tests were on and out of 3 options only 1 matched the row. For now we simply don't generate if the row breaks. Need to discuss what to do if the example exists but the can't be used by some option in the oneOf.
* Updated some language to say `specification` instead of `contract`.

**Why**:

* The bug fix was required as without that generative tests were not working. It's more of a temp fix really to make sure generative tests can at least run, until we decide what's the real fix.
* The language fix was needed as the older language was looking awkward in a recent video we recorded.

**How**:

Updated newBasedOn and negativeBasedOn in Anypattern, and some language strings where needed.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #123


harikrishnan83 (approved):
Looks good to me

PR #800, opened 2023-10-26
Fix the success() method in the Results class. issue #792 

### LINKED ISSUE #792: Fix .success() method in Results
**Description**
The Results.success() method returns true if there are successes, but does not check that there are no failures

It should return true if successes >1 and failures == 0.

Make sure the tests are passing, and if not please discuss the test failures with us.

TejasDhole:
**What**:Here i've Fix the `success()` method in the Results class.

**Why**:The `success()` method should return `true` if there are one or more successes and no failures. The current implementation does not check if there are no failures. This fix is necessary to correctly determine whether there are any failures.

**How**:I have updated the `success()` method in the `Results` class to check if there are successes (`Result.Success`) and no failures (`Result.Failure`). The method now returns `true` only when there are one or more successes and no failures.

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #792 

Test cases were getting failed in my system so as per repo maintainers suggest to create PR im doing this.

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Results.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Results.kt
@@ -6,7 +6,7 @@ data class Results(val results: List<Result> = emptyList()) {
     fun hasResults(): Boolean = results.isNotEmpty()
 
     fun hasFailures(): Boolean = results.any { it is Result.Failure }
-    fun success(): Boolean = !hasFailures()
+    fun success(): Boolean = successCount > 0 && failureCount == 0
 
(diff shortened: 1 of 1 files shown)


harikrishnan83:
@TejasDhole Thanks for the PR. Can you please let us know if you intend to fix the failing tests on this PR? Without that we will be unable to review your changes. cc @joelrosario

joelrosario:
Thanks for raising this PR @TejasDhole! Closing this, as we have merged #890, which includes a commit from you that addresses this issue.

PR #807, opened 2023-11-05
Contract test suites

joelrosario:
**What**:

The tests are now divided into suites. This is reflected in two ways:
1. When the tests are run using SpecmaticJUnitSupport, the tests are now grouped by suite instead of being in one long list.
2. The test names now have a prefix against the name reflecting the suite that they're in.

**Why**:

It became necessary to be able to tell a team, just pay attention to the "within-bounds" tests for now. It would be more convenient if we could give them a switch to avoid the noise.

**How**:

We've updated added a `--suite` parameter. It takes 'W' for Within-bounds tests, WG for Within-bounds generated tests, and O for Outside-bounds tests.

The test names show the suite name as a prefix to the test name if running from the terminal.

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #123


joelrosario:
Will redo this PR later

PR #820, opened 2023-11-17
allow setting OAuth2 token through environment variable

harikrishnan83:
**What**:

Externalise OAuth2 token to environment variable

**Why**:

We will not alway be able to pass the OAuth2 token through specmatic.json in usecases such as when specmatic is being used from command line 

**How**:

If OAuth2 token is not available in configuration pick it up from environment. This is assuming people who are setting this env variable will not have specatmatic.json and even if they do, have not configured security token.

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate


joelrosario (approved):
I'd suggest flipping order... let getBearerTokenFromEnvironment be called first, so the environment variable can take precedence. However approving so this can move ahead.

PR #835, opened 2023-11-26
Miscellaneous fixes

joelrosario:
**What**:

* Fix: schema examples should be ignored if they are objects
* Fix: oneOf at the top-level of the schema was not being handled properly
* Fix: when a key existed in two different places of the same request, generative tests sometimes generated an example that did not match the specification. This PR fixes the issue by improving how the REQUEST-BODY is used. It is not flattened now. Instead, the object is used directly. However the implementation does not now cut off generation when generative tests is switched on.
* Remove generativeTestsEnabled from Feature, Scenario, tests, making this flag a top-level chop.
* Made ONLY_POSITIVE a top-level flag

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


harikrishnan83 (approved):
Looks good to me

PR #846, opened 2023-12-06
#721 Implemented support for the RFC3339 datetime format.

### LINKED ISSUE #721: Date-time strings do not contain offset part
**Description**
As per https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 which is the notation mandated by openapi's `date-time` format, the date time strings should always contain timezone offset part. 
Specmatic generates date-time strings without offset information: e.g. `2023-09-01T14:37:25.347356`

**Steps to reproduce**
1. define a parameter of type date-time:
```yaml
schema:
  type: string
  format: date-time
```
2.run tests

**Expected behavior**
A valid RFC 3339 date-time string should be generated (for example by appending `Z` at the end or by appending time offset based on current TZ)

jaydeepk:
**What**:
Implemented support for the RFC3339 datetime format as defined in the [open api specification](https://swagger.io/docs/specification/data-models/data-types/#string)
This means that:
- The datetime strings generated by specmatic will include the  timezone offset as specified in [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) 
- All datetime strings sent to specmatic ( as part of examples, stub requests) would need to confirm to the [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) format.

**Checklist**:
- [x] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #721

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/pattern/DatePattern.kt
+++ b/core/src/main/kotlin/in/specmatic/core/pattern/DatePattern.kt
@@ -5,7 +5,6 @@ import `in`.specmatic.core.Result
 import `in`.specmatic.core.value.JSONArrayValue
 import `in`.specmatic.core.value.StringValue
 import `in`.specmatic.core.value.Value
-import java.time.LocalDateTime
 
 object DatePattern : Pattern, ScalarType {
     override fun matches(sampleData: Value?, resolver: Resolver): Result = when (sampleData) {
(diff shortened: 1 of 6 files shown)


harikrishnan83 (approved):
Looks good to me

joelrosario (approved):
LGTM

PR #854, opened 2023-12-08
Support json and yml extensions

### LINKED ISSUE #848: Failing to use with simple example/plain text responses
Just downloaded Specmatic and tried to get it working on a super-simple microservices example I've used with other tools. But two things so far didn't work as expected:

1. It didn't like my openapi.json file, I had to convert it to yaml before I could do anything. Are there any plans to support the json format also, as many tools (such as Python flask which is my example technology) make this format the default?
2. My endpoint returns plain text, and the openapi spec clearly says that it will do this. Specmatic however, wanted json responses anyway and failed with:
Contract expected JSON object but response contained "Hello, GMUOS!" 

```
    responses:
        '200':
          content:
            text/html:
              schema:
                $ref: '#/components/schemas/Generated'
          description: Successful response
```
I realise not many real applications expect plain text, but shouldn't it be allowed? It seems like Specmatic is ignoring this aspect of the spec.

Response at 2023-12-7 3:31:6.739
    200 OK
    Server: Werkzeug/2.3.7 Python/3.10.6
    Date: Thu, 07 Dec 2023 14:31:06 GMT
    Content-Type: text/html; charset=utf-8
    Content-Length: 13
    Access-Control-Allow-Origin: *
    Connection: close

    Hello, GMUOS!

harikrishnan83:
**What**:

Equivalent YAML and JSON representations of the OpenAPI should be acceptable.

**Why**:

Removes the need for converting JSON to YAML specifications.

**How**:

Updating supported extensions list (underlying parser already supported it).

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

**Issue ID**:
Closes: #848


joelrosario (approved):
LGTM

PR #868, opened 2023-12-13
When loading example as stub, use a fake auth parameter

joelrosario:
**What**:

When loading an expectation from an example in the specification, if a security scheme was provided, the expectation would be rejected with an error saying the respective header or query parameter was missing.

**Why**:

Security scheme examples can't be provided in the specification, so examples of APIs with security schemes can't become valid expectations without a little help from Specmatic.

**How**:

Specmatic adds a dummy security header / query parameter in the expectation when it comes from an example, which lets the expectation match requests that have the security parameter. The value is not validated in any case (except for bearer auth and only to ensure that it starts with bearer).

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

**Issue ID**:
Closes: #123


harikrishnan83 (approved):
Looks good to me

PR #886, opened 2023-12-24
Log request content-type in contract test

harikrishnan83:
**What**:

Log request content-type header for contract tests

**Why**:

* Current implementation identifies request content-type based on the content being sent and not the what is being mentioned in the API specification
* Without logs that show the content-type we are not even aware when this is happening

**How**:

* Get the request header from KtorRequest and set it on Specmatic HttpRequest so that it gets logged

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate


joelrosario (approved):
Looks good to me.

PR #890, opened 2023-12-28
Issue 792

### LINKED ISSUE #792: Fix .success() method in Results

v1bh0r:
What: Here i've Fix the success() method in the Results class. I have piggy-backed on the PR created by Tejas: https://github.com/znsio/specmatic/pull/800

Why: The success() method should return true if there are one or more successes and no failures. The current implementation does not check if there are no failures. This fix is necessary to correctly determine whether there are any failures.

How: I have updated the success() method in the Results class to check if there are successes (Result.Success) and no failures (Result.Failure). The method now returns true only when there are one or more successes and no failures. However, the result is success if there are 0 results.

Checklist:

 Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation
 Tests
 Sonar Quality Gate
Issue ID:
Closes: https://github.com/znsio/specmatic/issues/792

### DIFF
--- a/application/src/main/kotlin/application/CompatibleCommand.kt
+++ b/application/src/main/kotlin/application/CompatibleCommand.kt
@@ -360,8 +360,8 @@ internal data class CompatibilityOutput(val exitCode: Int, val message: String)
 internal fun compatibilityMessage(results: Outcome<Results>): CompatibilityOutput {
     return when {
         results.result == null -> CompatibilityOutput(1, results.errorMessage)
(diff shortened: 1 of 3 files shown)


joelrosario (approved):
Thanks for the PR, @v1bh0r and @TejasDhole!

I've added a commit of my own as well.

I'm going ahead and merging it.

PR #891, opened 2023-12-29
Api Coverage Report improvements

jaydeepk:
**What**:
Api Coverage Report improvements:
- An endpoint will be reported as covered if the corresponding contract test fails and the actuator is not available (endpoints api is not set)
- Used the spec endpoints to determine 'missing in spec' endpoints rather than using contract test results.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


joelrosario (approved):
Looks good to me.

PR #931, opened 2024-01-22
Add test showing the incorrect behaviour when two or more headers required

mariusz-mc:


ketan:
Closing since we don't have any fixes for the issue.

PR #1058, opened 2024-04-06
Removed unnecessary jvmTarget 17 in kotlinOptions from all build.gradle files

mileejain:
**What**:

Removed unnecessary jvmTarget 17 in kotlinOptions from all build.gradle files.
In some files this was even mentioned multiple times. Which is redundant.

**Why**:

It is redundant and confusing to set jvmTarget to 17

**How**:

**Checklist**:
- [x] Tests

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -114,12 +114,6 @@ jacocoTestReport {
 
 group = 'in.specmatic'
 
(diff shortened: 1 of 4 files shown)


joelrosario (approved):
Looks good to me, thanks @mileejain!

PR #1059, opened 2024-04-08
Refactored Test for Socket Timeout Issue

burdeamit:
**What**:  
- This PR includes refactoring the `Socket Timeout Test` by making use of `ApacheHttpClientFactory` to create a new HttpClient. 
- This PR is refactoring the test added in PR #1053

**Why**: 
The purpose of these changes is to ensure that the `ApacheHttpClientFactory` is correctly setting the timeout policy when creating a new `HttpClient`. This is crucial for the correct operation of the HTTP client in scenarios where network timeouts can occur.

**How**:

 - Added test case where the client should set a timeout policy with `socketTimeout` giving breathing room for `requestTimeout` to kick in first. 
 - This test creates an instance of `ApacheHttpClientFactory` with a random timeout value and then checks that the `requestTimeoutInMillis` and `socketTimeoutInMillis` properties of the `timeoutPolicy` from the factory match the expected values.  

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

### DIFF
--- a/application/src/test/kotlin/in/specmatic/test/ApacheHttpClientFactoryTest.kt
+++ b/application/src/test/kotlin/in/specmatic/test/ApacheHttpClientFactoryTest.kt
@@ -1,6 +1,5 @@
 package `in`.specmatic.test
 
-import io.mockk.every
 import io.mockk.justRun
 import io.mockk.mockk
 import io.mockk.verify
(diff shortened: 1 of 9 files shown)


joelrosario (approved):
Looks good to me. Thanks @burdeamit!

PR #1081, opened 2024-04-29
Update the pattern used to parse RFC3339 compliant datetime

### LINKED ISSUE #1078: Specmatic unable to parse valid DateTime
**Description**
I am running Contract Tests with Specmatic against a Spring Boot app. My OpenAPI spec has the field defined as:

```
        createDatetime:
          type: string
          format: date-time
```

In my Java model the field is:

`private OffsetDateTime createDatetime;`

When my contract tests run, the API produces a response of `"createDatetime": "2024-04-25T09:06:26.572577Z",` which is a valid ISO 8601 format.

However, Specmatic produces the following error:

```
	  >> RESPONSE.BODY.createDatetime
	  
	     Error: Text '2024-04-25T09:06:26.572577Z' could not be parsed at index 19
```

It doesn't seem to like the `Z`.

**Steps to reproduce**
See Description

**Expected behavior**
Specmatic should accept `2024-04-25T09:06:26.572577Z` as a valid datetime.

**Screenshots**
N/A

**System Information:**
 - Specmatic version: 1.3.11
 - JDK version: 21

**Additional context**
Related issue - https://github.com/znsio/specmatic/issues/721

yogeshnikam671:
**What**:
Fixes the parsing of RFC3339 compliant datetime values.

**Why**:
The existing parsing logic of RFC3339 compliant datetime values used to fail for certain combinations. 
e.g. 2024-04-25T09:06:26.572577123456Z
This fix will make sure that the parsing succeeds for all those combinations.

**How**:
This is fixed by updating the pattern used to match the RFC3339 compliant strings.

**Checklist**:
- [x] Tests
- [x] Sonar Quality Gate

**Issue ID**:
Closes: #1078

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/pattern/RFC3339.kt
+++ b/core/src/main/kotlin/in/specmatic/core/pattern/RFC3339.kt
@@ -4,18 +4,31 @@ import java.time.LocalDateTime
 import java.time.ZoneId
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
+import java.util.regex.Pattern
 
 private const val DATETIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ssXXX"
 private const val DATE_FORMAT = "yyyy-MM-dd"
 
+private val RFC3339_PATTERN: Pattern = Pattern.compile (
+    "^(\\d{4})-(\\d{2})-(\\d{2})" // yyyy-MM-dd
+            + "([Tt](\\d{2}):(\\d{2}):(\\d{2})(\\.\\d+)?)?" // 'T'HH:mm:ss.milliseconds
+            + "([Zz]|([+-])(\\d{2}):(\\d{2}))?" // 'Z' or time zone shift HH:mm following '+' or '-'
+)
+
 class RFC3339 {
     companion object{
-        val dateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern(DATETIME_FORMAT)
         val dateFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern(DATE_FORMAT)
+
+        fun parse(dateTime: String) {
+            val matcher = RFC3339_PATTERN.matcher(dateTime)
+            if(!matcher.matches()) {
+                throw ContractException("Error while parsing the dateTime as per RFC 3339: $dateTime")
+            }
+        }
 
         fun currentDateTime(): String {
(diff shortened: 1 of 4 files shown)


core/src/test/kotlin/in/specmatic/core/pattern/DateTimePatternTest.kt
```diff
@@ -48,14 +48,20 @@ internal class DateTimePatternTest {
 
     @Test
     fun `should match RFC3339 date time format`() {
```
harikrishnan83: Should convert this to a parameterised test? and also have one or two tests for non RFC 3339 compliant dates so that we can verify the parse fails?

core/src/main/kotlin/in/specmatic/core/pattern/RFC3339.kt
```diff
@@ -4,18 +4,31 @@ import java.time.LocalDateTime
...
+
 class RFC3339 {
     companion object{
-        val dateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern(DATETIME_FORMAT)
         val dateFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern(DATE_FORMAT)
 
+        fun parse(dateTime: String) {
+            val matcher = RFC3339_PATTERN.matcher(dateTime)
+            if(!matcher.matches()) {
+                throw ContractException("Error while parsing the dateTime: $dateTime")
```
harikrishnan83: Is there any match error message that can be appended to this Exception? Will help to say why the parse failed.

harikrishnan83 (changes requested):
Looks good, please go through some of the minor comments I have added and then merge the PR.

core/src/main/kotlin/in/specmatic/core/pattern/RFC3339.kt
yogeshnikam671: I did not find one. Seems like there isn't one.

core/src/test/kotlin/in/specmatic/core/pattern/DateTimePatternTest.kt
yogeshnikam671: Pushed the relevant changes. Build is in progress.

joelrosario (approved):
Looks good to me!

harikrishnan83 (approved):

PR #1082, opened 2024-04-29
Add indexes to test names so that Specmatic always generates tests with unique names.

burdeamit:
**What**:

At present, Specmatic adds an index suffix to all negative tests with the same name, however this behaviour is not implemented for positive tests. This results in Specmatic generating positive tests with the same name duplicated.
This PR enforces a consistent behaviour so that, whenever a group of tests [ Positive or Negative ] with the same name are generated, an index is suffixed to the test name so that all tests have unique names.

**Why**:

Duplicate test names result in python wrapper reporting an incorrect test count. 

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/Feature.kt
+++ b/core/src/main/kotlin/in/specmatic/core/Feature.kt
@@ -353,7 +353,7 @@ data class Feature(
             else
                 flagsBased.withoutGenerativeTests()
             val testScenarios =  originalScenario.generateTestScenarios(resolverStrategies, testVariables, testBaseURLs)
-            addIndexes(originalScenario, testScenarios, flagsBased.positivePrefix)
+            addIndexAndPrefixToDistinguishTests(originalScenario, testScenarios, flagsBased.positivePrefix)
         }
 
     fun negativeTestScenarios(): Sequence<Pair<Scenario, ReturnValue<Scenario>>> {
@@ -371,7 +371,7 @@ data class Feature(
                     originalScenario.httpRequestPattern.matches(sampleRequest, originalScenario.resolver).isSuccess()
                 }
             }
-            addIndexes(negativeScenario, nonMatchingNegativeTestScenarios, flagsBased.negativePrefix)
+            addIndexAndPrefixToDistinguishTests(negativeScenario, nonMatchingNegativeTestScenarios, flagsBased.negativePrefix)
         }
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/in/specmatic/core/Feature.kt
```diff
@@ -365,19 +365,13 @@ data class Feature(
             val negativeTestScenarios =
                 negativeScenario.generateTestScenarios(flagsBased, testVariables, testBaseURLs)
 
-            negativeTestScenarios.filterNot { negativeTestScenarioR ->
+            val filtered = negativeTestScenarios.filterNot { negativeTestScenarioR ->
```
harikrishnan83: Can we please rename `filtered` to what it actually contains? This will help readability of line where `filtered` is being used?

harikrishnan83: not sure I understand why we are naming this `negativeTestScenarioR`? what is the `R`? Can you please check?

core/src/main/kotlin/in/specmatic/core/Feature.kt
```diff
@@ -400,6 +394,27 @@ data class Feature(
         serverState = emptyMap()
     }
 
+    private fun addIndexes(
```
harikrishnan83: Technically this method is adding both `+v`/`-ve` prefix as well as index, should it be named appropriately?

harikrishnan83: Will also be super helpful to unit test this method separately. While the feature level tests are very useful, this will allow is try all difficult combinations.

harikrishnan83: Is this method doing too much, should we separate out application of prefix and index as two separate `collections.map` operations?

core/src/test/kotlin/in/specmatic/core/FeatureTest.kt
```diff
@@ -1482,7 +1482,7 @@ paths:
     }
 
     @Test
-    fun `positive examples of 4xx should be able to have non-string non-spec-conformant examples`() {
```
harikrishnan83: Can we please restore this older test that got accidentally removed?

harikrishnan83 (changes requested):
Thanks, requires minor changes.

core/src/main/kotlin/in/specmatic/core/Feature.kt
burdeamit: -  `negativeTestScenarioR` is essentially a temporary variable used to hold each individual negative test scenario as it's being evaluated by the `filterNot` function. 
- This will be renamed to `currentNegativeTestScenario`.  So the name indicates that this is the current negative test scenario being evaluated in the filter operation.

burdeamit: Fixed in:  [Renamed the variable in negativeTestScenarios fun to more descriptive…](https://github.com/znsio/specmatic/pull/1082/commits/93d1ade4e6761380c18b777bfe94ab1eca681531)

burdeamit: Fixed in : [Renamed variable 'negativeTestScenarioR' to 'currentNegativeTestScena…](https://github.com/znsio/specmatic/pull/1082/commits/c0793823051441f79568451ee8e85e9d7bd588cd)

core/src/test/kotlin/in/specmatic/core/FeatureTest.kt
burdeamit: Fixed in: [Restored the test: positive examples of 4xx should be able to have no…](https://github.com/znsio/specmatic/pull/1082/commits/b5db049d8b9d7d80cef17c55b9b95c915937466a)

core/src/main/kotlin/in/specmatic/core/Feature.kt
burdeamit: ###  ToDo's
- [x] Technically this method adds both the +v/-ve prefix and the index, should it be named appropriately?
- [ ] It will also be super helpful to unit test this method separately. While the feature-level tests are very useful, this will allow us to try all difficult combinations.
- [ ] Is this method doing too much, should we separate the application of prefix and index as two separate collections.map operations?

joelrosario:
Closing this as we are using the description of the test as a unique tag instead of the number.

PR #1130, opened 2024-05-28
Two seconds default grace period and timeout for HttpStub

yogeshnikam671:
**What**:
The contract test were passing but in the logs the following error was observed -
```
io.netty.channel.AbstractChannelHandlerContext invokeExceptionCaught
WARNING: Failed to submit an exceptionCaught() event.
java.util.concurrent.RejectedExecutionException: event executor terminated 
```
This PR introduces a fix which gets rid of such errors.

**Why**:
These changes are necessary to ensure graceful shutdown of HttpStub. 
And also to provide cleaner test output logs.

**How**:
We use Netty server as a HttpStub. The stop method of Netty server expects two arguments.
1. **gracePeriodMillis** - Time for which the server waits for the ongoing tasks' completion before terminating them.
2. **timeoutMillis** - Time for which the server waits for tasks to get completely terminated before shutting itself down.

We were setting the `gracePeriodMillis` to `0` by default. Because of this, the ongoing tasks were getting interrupted leading to the errors in the test output logs.

**The fix :**
1. Change the default `gracePeriodMillis` from `0` to `2000` while creating the stub using `createStub` APIs

**Checklist**:

- [x] Tests
- [x] Sonar Quality Gate

### DIFF
--- a/application/src/main/kotlin/application/HTTPStubEngine.kt
+++ b/application/src/main/kotlin/application/HTTPStubEngine.kt
@@ -41,7 +41,8 @@ class HTTPStubEngine {
             passThroughTargetBase = passThroughTargetBase,
             httpClientFactory = httpClientFactory,
             workingDirectory = workingDirectory,
-            specmaticConfigPath = specmaticConfigPath
+            specmaticConfigPath = specmaticConfigPath,
+            timeoutMillis = 2000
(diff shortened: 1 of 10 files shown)


joelrosario (approved):
LGTM

PR #1139, opened 2024-06-08
Fail when API specs mentioned in specmatic.json are not found

harikrishnan83:
Fail when API specs mentioned in specmatic.json are not found


joelrosario (approved):
LGTM

PR #1140, opened 2024-06-09
Log error when invalid contract paths are selected by user

yogeshnikam671:
**What**:
If we run the test against a folder (i.e. not a valid contract file path),
we get following log messages -
```
Endpoints API not found, cannot calculate actual coverage

Could not load report configuration, coverage will be calculated but no coverage threshold will be enforced
The Open API coverage report generated is blank.
This can happen if you have included all the endpoints in the 'excludedEndpoints' array in the report section in specmatic.json, or if your open api specification does not have any paths documented.
```

It does not mention that it is failing because a file with valid extension was not selected.

With this change, we will get additional error message as follows -
```
Endpoints API not found, cannot calculate actual coverage

Could not load report configuration, coverage will be calculated but no coverage threshold will be enforced
The Open API coverage report generated is blank.
This can happen if you have included all the endpoints in the 'excludedEndpoints' array in the report section in specmatic.json, or if your open api specification does not have any paths documented.

No valid specification file found at given paths : /Users/yogeshanandanikam/personal_projects/specmatic-intellij-plugin
```

**Why**:
Better user experience.

**How**:
Logs a message when the contract paths are not empty but none of them is valid.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/stub/api.kt
+++ b/core/src/main/kotlin/in/specmatic/stub/api.kt
@@ -6,6 +6,7 @@ import `in`.specmatic.core.git.SystemGit
 import `in`.specmatic.core.log.StringLog
 import `in`.specmatic.core.log.consoleLog
 import `in`.specmatic.core.log.logger
+import `in`.specmatic.core.pattern.ContractException
 import `in`.specmatic.core.utilities.ContractPathData
 import `in`.specmatic.core.utilities.contractStubPaths
 import `in`.specmatic.core.utilities.examplesDirFor
@@ -70,6 +71,13 @@ fun createStubFromContracts(
     host: String = "localhost",
     port: Int = 9000
 ): ContractStub {
+    val validContractPaths = contractPaths.filter { path ->
+        File(path).extension in CONTRACT_EXTENSIONS
+    }
+    if (validContractPaths.isEmpty()) {
+        throw ContractException("No valid specification file found at given paths : $contractPaths")
+    }
+
     return createStubFromContracts(
         contractPaths,
         dataDirPaths,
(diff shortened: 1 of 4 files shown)


harikrishnan83 (changes requested):
Should we do this for stub?

core/src/main/kotlin/in/specmatic/stub/api.kt
```diff
@@ -70,6 +71,13 @@ fun createStubFromContracts(
     host: String = "localhost",
     port: Int = 9000
 ): ContractStub {
+    val validContractPaths = contractPaths.filter { path ->
```
joelrosario: The logic for filtering is duplicated 3 times. Can you pull it out into a common function?

core/src/main/kotlin/in/specmatic/stub/api.kt
```diff
@@ -70,6 +71,13 @@ fun createStubFromContracts(
...
+        File(path).extension in CONTRACT_EXTENSIONS
+    }
+    if (validContractPaths.isEmpty()) {
+        throw ContractException("No valid specification file found at given paths : $contractPaths")
```
joelrosario: contractPaths here is a list. Can you confirm that it appears as a CSV when the error message is printed?

The same applies to the change to createStubFromContracts on line 96 in this file.

joelrosario (commented):
Hi @yogeshnikam671, thanks, the logic looks good. I've requested a few changes with respect to removing duplication and the format of the error message.

joelrosario (approved):
LGTM

PR #1142, opened 2024-06-12
Add support for specmatic configuration in yaml format

yogeshnikam671:
Add support for Specmatic configuration in yaml format.


harikrishnan83 (approved):
Looks good to me

PR #1160, opened 2024-07-01
Fix bug in generate method of NumberPattern where it was generating d…

yogeshnikam671:
**What**
Fixes bug in generate method of NumberPattern where it is generating double values even though isDoubleFormat flag is false.


joelrosario (approved):
LGTM

PR #1162, opened 2024-07-02
Use `_examples` as default folder for both externalised test and stub data

harikrishnan83:
**What**:

Use `_examples` as default folder for both externalised test and stub data. We are also maintaining backward compatibility for `_data` and `_tests` for stub and test data respectively

Because of these above changes, bumping up minor version (from 3 to 4).

**Why**:

Currently we pick up stub data from `_data` and test data from `_tests`. Since both test and stub data format is the same, it will be easier if we can keep it under one dir. This way we not have to copy over the files from one to the other to run them as tests etc.

**How**:

* Renamed default test data dir to `_tests` from `_examples`.
* Renamed default test data dir to `_data` from `_examples`.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


joelrosario (approved):
LGTM

PR #1176, opened 2024-07-10
Enable generation of double value when isDoubleFormat is true and min and max are not set

yogeshnikam671:
**What**:
NumberPattern was not generating a double value when min and max were not set even though isDoubleFormat is true.

**Why**:
It was a missing scenario.

**How**:
Fixed it by explicitly adding a check for isDoubleFormat in NumberPattern's generate method.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/pattern/NumberPattern.kt
+++ b/core/src/main/kotlin/in/specmatic/core/pattern/NumberPattern.kt
@@ -19,7 +19,7 @@ data class NumberPattern(
     val maximum: BigDecimal = HIGHEST_DECIMAL,
     val exclusiveMaximum: Boolean = false,
     override val example: String? = null,
-    val isDoubleFormat: Boolean = true
+    val isDoubleFormat: Boolean = false
 ) : Pattern, ScalarType, HasDefaultExample {
 
   companion object {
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/in/specmatic/core/pattern/NumberPattern.kt
```diff
@@ -19,7 +19,7 @@ data class NumberPattern(
     val maximum: BigDecimal = HIGHEST_DECIMAL,
     val exclusiveMaximum: Boolean = false,
     override val example: String? = null,
-    val isDoubleFormat: Boolean = true
+    val isDoubleFormat: Boolean = false
```
joelrosario, line 22: Why was the default changed from true to false?

joelrosario (changes requested):
Mostly looks good, but I am not sure why the default was switched from true to false. Would be good to keep the default as-is, unless there's a reason to change it.

core/src/main/kotlin/in/specmatic/core/pattern/NumberPattern.kt
yogeshnikam671: When `isDoubleFormat` was introduced, its value was kept as true by default. 
This default value was working until now since we were **not returning a double value** when min and max were not set. 
Now that we are returning a double value in this scenario, the tests started failing where we were expecting a non-double (Int) value.
Hence, I changed the default value which passed those tests and also maintained the behaviour where if we don't pass the isDoubleFormat value as true, we will get Int value back.

joelrosario (approved):
LGTM

PR #1178, opened 2024-07-11
Revert invalid contract path check

yogeshnikam671:
**What**:
Revert the check where we were throwing an exception if all contract paths are invalid.

**Why**:
This change has some bug which needs some investigation to resolve.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate
- [x] Tested against sample projects


joelrosario (approved):
LGTM

PR #1179, opened 2024-07-12
Enable negativeBasedOn to generate negatives based on the config passed.

yogeshnikam671:
Currently 'withDataTypesNegatives' option is supported where negativeBasedOn will exclude data type based negatives.

**What**:
Added an argument `config` in the `negativeBasedOn` function which enables to generate different variations of negative patterns to be generated by the function.

**Why**:
Need this for a scenario where only constraint based negatives are required.

**Checklist**:

- [x] Tests
- [x] Tested against sample project
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/HttpPathPattern.kt
+++ b/core/src/main/kotlin/in/specmatic/core/HttpPathPattern.kt
@@ -157,7 +157,10 @@ data class HttpPathPattern(
     private fun negatively(patterns: List<URLPathSegmentPattern>, row: Row, resolver: Resolver): Sequence<ReturnValue<List<URLPathSegmentPattern>>> {
         val current = patterns.firstOrNull() ?: return emptySequence()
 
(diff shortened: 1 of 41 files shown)


core/src/main/kotlin/in/specmatic/core/pattern/models/NegativePatternConfiguration.kt
```diff
@@ -0,0 +1,5 @@
+package `in`.specmatic.core.pattern.models
```
joelrosario: instead of `pattern.models`, can you call it `pattern.config`.

joelrosario (changes requested):
Mostly good. Just one comment. Also can you merge with main, as there may be a few conflicts.

core/src/main/kotlin/in/specmatic/core/pattern/models/NegativePatternConfiguration.kt
yogeshnikam671: done

joelrosario (approved):
LGTM

PR #1184, opened 2024-07-17
[BUG FIX] breadCrumb key getting printed even though there is no message associated with it

yogeshnikam671:
**What**:
Fixes a minor bug in the `singleLineDescription` method of `ValueDetails` which was printing the breadCrumb key even though there is no message associated with it.

**How**:
Added a condition where we are returning the singleLineDescription as empty if there is no message associated with a breadCrumb key.

**Checklist**:

- [x] Tests
- [x] Tested against sample project
- [ ] Sonar Quality Gate


joelrosario (approved):
LGTM

PR #1185, opened 2024-07-17
Alias `test` to `provides` and `stub` to `consumes` for Specmatic Config.

StarKhan6368:
**What**: Alias `test` to `exposes` and `stub` to `depends`

**Why**: 

**How**:  Using the `@JsonAlias` annotation.

**Checklist**:

- [x] Documentation added to the README.md OR link to PR on https://github.com/specmatic/specmatic-documentation
- https://github.com/znsio/specmatic-documentation/pull/66
- [x] Tests
- [x] Tested with Sample Projects
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/in/specmatic/core/SpecmaticConfig.kt
+++ b/core/src/main/kotlin/in/specmatic/core/SpecmaticConfig.kt
@@ -75,7 +75,7 @@ data class Source(
     val branch: String? = null,
     @field:JsonAlias("exposes")
     val test: List<String>? = null,
-    @field:JsonAlias("depends")
+    @field:JsonAlias("dependencies")
     val stub: List<String>? = null,
     val directory: String? = null,
 )
--- a/core/src/main/kotlin/in/specmatic/core/SpecmaticConfig.kt
+++ b/core/src/main/kotlin/in/specmatic/core/SpecmaticConfig.kt
@@ -1,9 +1,6 @@
 package `in`.specmatic.core
 
-import com.fasterxml.jackson.annotation.JsonProperty
-import com.fasterxml.jackson.annotation.JsonSubTypes
-import com.fasterxml.jackson.annotation.JsonTypeInfo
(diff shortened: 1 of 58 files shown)


core/src/main/kotlin/in/specmatic/core/SpecmaticConfig.kt
```diff
@@ -76,7 +73,9 @@ data class Source(
     val provider: SourceProvider = SourceProvider.filesystem,
     val repository: String? = null,
     val branch: String? = null,
+    @field:JsonAlias("exposes")
     val test: List<String>? = null,
+    @field:JsonAlias("depends")
```
joelrosario: This should be `dependencies`

joelrosario (changes requested):
@StarKhan6368 can you test these in the sample projects as well by updating specmatic.yml / json to use dependencies and exposes (both specmatic-order-bff-java and specmatic-order-api-java)

core/src/test/resources/specmaticConfigFiles/specmatic_exp_dep.json
```diff
@@ -0,0 +1,83 @@
+{
+  "sources": [
+    {
+      "provider": "git",
+      "repository": "https://contracts",
+      "exposes": [
+        "com/petstore/1.spec"
```
joelrosario: Can you use yaml instead of spec for the extensions, as .spec is the legacy format.

joelrosario (changes requested):
I've added a comment requesting changes.

core/src/test/resources/specmaticConfigFiles/specmatic_exp_dep.json
StarKhan6368: Renamed OAS file extensions to .yaml instead of legacy .spec, updated the test expectations as well.

joelrosario (approved):
LGTM

PR #1202, opened 2024-07-23
Support configuration of some flags through Specmatic configuration file

yogeshnikam671:
Enable support for following flags to be configured via the configuration file - 

1. SPECMATIC_GENERATIVE_TESTS as 'enableResiliencyTests' 
2. ONLY_POSITIVE as 'enableOnlyPositiveTests'
3.  VALIDATE_RESPONSE_VALUE as 'enableResponseValueValidation'
4.  EXTENSIBLE_SCHEMA as 'enableExtensibleSchema' 
5. MAX_TEST_REQUEST_COMBINATIONS as 'maxTestRequestCombinations' 
6. SCHEMA_EXAMPLE_DEFAULT as 'schemaExampleDefault'

**Checklist**:

- [x] Documentation (https://github.com/znsio/specmatic-documentation/pull/69)
- [x] Tests
- [ ] Sonar Quality Gate


harikrishnan83:
Should we group resiliencyTests parameters as a group?
```
resiliencyTests:
  enable: true
  onlyPositiveTests: true
  maxTestRequestCombinations: 10
enableExtensibleSchema: true
enableResponseValueValidaton:  true
```

harikrishnan83:
Can we add a section for `httpStub`? Mainly aspects like strict mode are hard to enable programmatically
```
proxy:
  httpStub:
    host: "localhost"
    port: "9000"
    baseURL: http://localhost:9000
    strict: true
    ssl: true
    examplesDir: './src/test/resources'
  graphql:
    ...
  grprc:
   ...
  kafkaMock:
   ...
```

alternatively,

```
stub:
  http:
    host: "localhost"
    port: "9000"
    baseURL: http://localhost:9000
    strict: true
    ssl: true
    examplesDir: './src/test/resources'
  graphql:
    ...
  grprc:
    ...

mock:
  kafka:
    ...
  jdbc:
    ...
```

core/src/main/kotlin/io/specmatic/conversions/EnvironmentAndPropertiesConfiguration.kt
```diff
@@ -1,62 +1,70 @@
 package io.specmatic.conversions
 
+import io.specmatic.core.SpecmaticConfig
 import org.apache.commons.lang3.BooleanUtils
 
-data class EnvironmentAndPropertiesConfiguration(val environmentVariables: Map<String, String>, val systemProperties: Map<Any?, Any?>) {
-    constructor() : this(System.getenv(), System.getProperties().toMap())
+data class EnvironmentAndPropertiesConfiguration(
```
joelrosario: The name of this class no more describes what it's doing. It reads all Specmatic configuration, which includes the config file, properties and environment variables.

How about renaming it to `Configuration`?

Note: there already is another class named `Configuration`, which may be more appropriately renamed `ConfigurationConstants`. In fact, there are methods and properties in this class related to the Specmatic config file, which ideally could move to SpecmaticConfig. The `config` and `_config` properties could be moved easily into `SpecmaticConfig`, as they are referred to in just a few places. The variables named `innerGlobalConfigFileName`,  methods `getGlobalConfigFileName`, `getConfigFileName`, and property `globalConfigFileName` could be moved later on as that moving them may be a bit more complicated.

joelrosario (changes requested):
My comment primarily addresses the evolution of the purpose of this class, and proposes a few changes in line with this evolution.

core/src/test/kotlin/io/specmatic/conversions/OpenApiIntegrationTest.kt
```diff
@@ -156,8 +155,8 @@ Examples:
 
             @Test
             fun `should match http request for spec with oauth2 security scheme with token defined in environment variable even if the token value in the request is different`() {
-                val environmentAndPropertiesConfiguration = EnvironmentAndPropertiesConfiguration(mapOf("oAuth2AuthCode" to "ENV1234"), emptyMap())
-                val feature = parseContractFileToFeature("src/test/resources/openapi/hello_with_oauth2_authorization_code_flow.yaml", environmentAndPropertiesConfiguration = environmentAndPropertiesConfiguration)
+                System.setProperty("oAuth2AuthCode", "ENV1234")
```
joelrosario: If this test fails, then oAuth2AuthCode remains set, and other tests may get the flag and fail as a result. The earlier approach provided a way to encapsulate configuration so that Feature would just get it from an object. 

Two options here:
1. Use a "finally" block to clear this up (I prefer finally as what is being cleared is local to the test)
2. Introduce some other way to inject environment variables / system properties.

joelrosario (changes requested):

core/src/test/kotlin/io/specmatic/conversions/OpenApiIntegrationTest.kt
yogeshnikam671: Resolved using the first approach.

core/src/test/resources/specmaticConfigFiles/specmatic.yml
```diff
@@ -55,4 +55,11 @@ security:
...
       BasicAuth:
         type: basicAuth
-        token: Abc123
\ No newline at end of file
+        token: Abc123
+
+resiliencyTests:
+  enable: all
+  maxTestRequestCombinations: 50
+enableResponseValueValidation: true
```
joelrosario: Did we take a decision to retain the `enable` language? cc @harikrishnan83

core/src/test/resources/specmaticConfigFiles/specmatic.yml
```diff
@@ -55,4 +55,11 @@ security:
...
+enableExtensibleSchema: true
+enableSchemaExampleDefault: false
```
joelrosario: This is a mechanical name. Should we come up with alternative, e.g. `useSchemaDefaultInTests` cc @harikrishnan83

joelrosario (changes requested):

core/src/test/resources/specmaticConfigFiles/specmatic.yml
yogeshnikam671: This makes sense. Will update the name.

harikrishnan83 (approved):
Looks good to me. Please address the minor comments we discussed. Thanks.

PR #1204, opened 2024-07-26
Added example directories parameter to TestCommand and Specmatic Config, aliased in StubCommand.

StarKhan6368:
**What**:
-  Add `--examples` parameter to TestCommand, rename in StubCommand.
- Add `exampleDirectories` attribute to SpecmaticConfig.

**Why**: Currently it is not possible to configure the examples directory within the test command or SpecmaticConfig.

**How**:
- Alias `examples` to `--data` argument in `StubCommand`.
- Add `--examples` in `TestCommand`.
- Refactor `loadExternalisedExamples` to use list of directories.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/test/kotlin/integration_tests/LoadTestsFromExternalisedFiles.kt
+++ b/core/src/test/kotlin/integration_tests/LoadTestsFromExternalisedFiles.kt
@@ -10,7 +10,7 @@ import io.specmatic.core.SpecmaticConfig
 import io.specmatic.core.log.*
 import io.specmatic.core.pattern.parsedJSONArray
 import io.specmatic.core.pattern.parsedJSONObject
-import io.specmatic.core.utilities.Flags.Companion.LOCAL_TESTS_DIRECTORY
+import io.specmatic.core.utilities.Flags.Companion.EXAMPLE_DIRECTORIES
 import io.specmatic.core.value.JSONObjectValue
 import io.specmatic.core.value.Value
 import io.specmatic.test.TestExecutor
@@ -205,7 +205,7 @@ class LoadTestsFromExternalisedFiles {
                       description: Name of the person
         """.trimIndent()
 
-        System.setProperty(LOCAL_TESTS_DIRECTORY, "src/test/resources/local_tests")
+        System.setProperty(EXAMPLE_DIRECTORIES, "src/test/resources/local_tests")
         val feature = OpenApiSpecification
             .fromYAML(
                 spec,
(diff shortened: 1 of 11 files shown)


core/src/test/kotlin/integration_tests/LoadTestsFromExternalisedFiles.kt
```diff
@@ -227,7 +227,7 @@ class LoadTestsFromExternalisedFiles {
 
         assertThat(results.successCount).isEqualTo(1)
         assertThat(results.success()).withFailMessage(results.report()).isTrue()
-        System.clearProperty(LOCAL_TESTS_DIRECTORY)
+        System.clearProperty(EXAMPLE_DIRECTORIES)
```
joelrosario: This should be in a finally block, both here and anywhere else this property is set. That way if the test fails, the property will get unset and will not impact any other tests.

core/src/test/kotlin/io/specmatic/core/SpecmaticConfigKtTest.kt
```diff
@@ -183,7 +188,8 @@ internal class SpecmaticConfigKtTest {
         val props = mapOf(
             SPECMATIC_GENERATIVE_TESTS to "false",
             VALIDATE_RESPONSE_VALUE to "false",
-            EXTENSIBLE_SCHEMA to "false"
+            EXTENSIBLE_SCHEMA to "false",
+            EXAMPLE_DIRECTORIES to "api/tests,api/stubs"
```
joelrosario: This was initially confusing as we typically won't pass test and stub directories together. This should have either two test directoreis or two stub directores.

joelrosario (changes requested):

core/src/test/kotlin/io/specmatic/core/SpecmaticConfigKtTest.kt
StarKhan6368: Thanks, I've renamed these folders to be generic to avoid this confusion.

core/src/test/kotlin/integration_tests/LoadTestsFromExternalisedFiles.kt
StarKhan6368: Thanks, added logic to clear system properties in test files.

joelrosario (approved):
LGTM

PR #1230, opened 2024-08-08
Implement a global configuration for the delay of stub responses.

StarKhan6368:
**What**: 
1. Implement a global configuration for the delay of stub responses.
2. Ensure all delays are specified in milliseconds, but maintain backward compatibility for delays given in seconds within the example JSON.

**Why**: It is currently not possible to configure a global delay for all responses for stub; the only method is to define the delay within the example JSON.

**How**:
1. Introduce a new key in the `SpecmaticConfig` stub configuration.
2. Add a new argument to the `StubCommand` to define the delay in milliseconds.
3. Add a new key in the example JSON to specify the delay in milliseconds.
4. Default to the configuration delay when responding if the example delay is not present.
5. Prioritize example-specific delay over the global delay.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


joelrosario (approved):
LGTM

core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -59,7 +61,10 @@ fun String.loadContract(): Feature {
     return parseContractFileToFeature(File(this))
 }
 
-data class StubConfiguration(val generative: Boolean = false)
+data class StubConfiguration(
+    val generative: Boolean? = false,
+    val delayInMilliseconds: Long? = getLongValue(SPECMATIC_STUB_DELAY)
```
joelrosario, line 66: Should this not be kebab case?

cc @harikrishnan83

joelrosario (changes requested):

joelrosario (approved):

PR #1231, opened 2024-08-09
Add a health check endpoint for proxy and stub server

yogeshnikam671:
**What**:
Exposes a health check endpoint for proxy and stub server.

**Why**:
Users can hit this endpoint to confirm where they can start interacting with the proxy or stub.

**How**:
Adding a health check endpoint for proxy and stub.

**P.S.** Currently we are just returning the "UP" status. I couldn't think of when we can send a "DOWN" status. Please let me know if you have any scenario in mind where we should return "DOWN" as status.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

Screenshot 2024-08-09 at 12 47 24 PM


harikrishnan83 (approved):
Looks good to me

PR #1266, opened 2024-09-02
Add dump endpoint for Specmatic proxy server

yogeshnikam671:
**What**:
Adds a `/_specmatic/proxy/dump` POST endpoint for Specmatic proxy server.

**Why**:
Currently the Specmatic proxy server needs to be shut down for dumping the generated spec and examples into the output directory.
With this endpoint, we need not have to do that. We can trigger this dump of generated spec and examples without having to stop the proxy server.

**How**:
By adding an endpoint in the proxy server which can be hit to trigger the generation of spec and examples in the output directory.

**Checklist**:

- [x] Documentation [Documentation PR](https://github.com/znsio/specmatic-documentation/pull/88)
- [x] Tests
- [x] Tested locally
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/io/specmatic/proxy/Proxy.kt
+++ b/core/src/main/kotlin/io/specmatic/proxy/Proxy.kt
@@ -1,12 +1,23 @@
 package io.specmatic.proxy
 
+import io.ktor.http.*
 import io.ktor.server.application.*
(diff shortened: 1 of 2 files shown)


harikrishnan83 (approved):
Looks good to me

PR #1311, opened 2024-09-19
Add support for "null" type in oneOf

yogeshnikam671:
If I run the spec containing the following schema as a stub -
```
schema:
  type: object
  properties:
    id:
      type: number
    type:
      oneOf:
        - type: string
          enum:
            - BASIC
            - PREMIUM
        - type: "null"
```
And, I try to hit this endpoint by passing the value for type as null i.e. `"type": null`, I get the following error -
```
Contract expected ("BASIC" or "PREMIUM") but request contained null
```

This PR fixes this issue and supports null values if `oneOf` the types is "null"

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
@@ -1428,7 +1428,7 @@ class OpenApiSpecification(
 
                     pattern
                 } else if (schema.oneOf != null) {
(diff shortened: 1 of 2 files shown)


joelrosario:
@yogeshnikam671  found a syntactically valid alternative:

```yaml
schema:
  type: object
  properties:
    id:
      type: number
    type:
      oneOf:
        - type: string
          enum:
            - BASIC
            - PREMIUM
        - type: string
          enum:
            - null
          nullable: true
```

Closing this PR.

PR #1314, opened 2024-09-20
Stub strict mode example loading fixes

yogeshnikam671:
**What**:
In strict mode, the stub shouldn't start if the examples directory is invalid in some way.

**Why**:
To get early feedback on the invalid examples.

**How**:
By exiting if -
1. The provided examples directory is invalid.
2. One of the examples in the examples directory is invalid.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/application/src/main/kotlin/application/StubLoaderEngine.kt
+++ b/application/src/main/kotlin/application/StubLoaderEngine.kt
@@ -13,7 +13,12 @@ import java.io.File
 
 @Component
 class StubLoaderEngine {
-    fun loadStubs(contractPathDataList: List<ContractPathData>, dataDirs: List<String>, specmaticConfigPath: String? = null): List<Pair<Feature, List<ScenarioStub>>> {
+    fun loadStubs(
+        contractPathDataList: List<ContractPathData>,
+        dataDirs: List<String>,
+        specmaticConfigPath: String? = null,
+        strictMode: Boolean
+    ): List<Pair<Feature, List<ScenarioStub>>> {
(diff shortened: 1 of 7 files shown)


application/src/main/kotlin/application/StubLoaderEngine.kt
```diff
@@ -24,7 +29,7 @@ class StubLoaderEngine {
 
         return when {
             dataDirs.isNotEmpty() -> {
-                loadContractStubsFromFiles(contractPathDataList, dataDirs, specmaticConfig)
+                loadContractStubsFromFiles(contractPathDataList, dataDirs, specmaticConfig, strictMode)
```
harikrishnan83, line 32: is Strict mode not part of SpecmaticConfig? if yes, then no need to add another param right? Please correct me if I am wrong.

core/src/main/kotlin/io/specmatic/stub/api.kt
```diff
@@ -477,3 +495,9 @@ fun isOpenAPI(path: String): Boolean =
         false
     }
 
+fun exitIfInvalidExamplesDirExists(exampleDirPaths: List<String>) {
```
harikrishnan83: sounds very similar to `exitIfDoesNotExist`, dir or file, can we not reuse the same function?

harikrishnan83 (changes requested):

harikrishnan83 (approved):
Looks good to me

PR #1315, opened 2024-09-24
Example generations using dictionary in both interactive and command-line modes.

StarKhan6368:
**What**: Better Example generations by utilizing dictionary in both interactive and command-line modes.

**Why**:  
Currently, examples are generated using random values, but now users can specify a dictionary for value substitution when generating examples.

**How**:
By leveraging the `substituteDictionaryValues` method in a Dictionary class, we can enable `forceSubstitution`, which replaces placeholder values with actual values from a user-provided dictionary.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/Dictionary.kt
+++ b/core/src/main/kotlin/io/specmatic/core/Dictionary.kt
@@ -5,7 +5,7 @@ import io.specmatic.core.value.JSONObjectValue
 import io.specmatic.core.value.StringValue
 import io.specmatic.core.value.Value
 
-class Dictionary(val map: Map<String, Value> = emptyMap()) {
+class Dictionary(private val map: Map<String, Value> = emptyMap()) {
     private fun substituteDictionaryValues(value: JSONArrayValue, paths: List<String> = emptyList(), forceSubstitution: Boolean = false): Value {
         val newList = value.list.mapIndexed { index, valueInArray ->
             val indexesToAdd = listOf("[$index]", "[*]")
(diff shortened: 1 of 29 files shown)


core/src/main/kotlin/io/specmatic/core/Substitution.kt
```diff
@@ -106,7 +106,7 @@ class Substitution(
     fun substituteSimpleVariableLookup(string: String, key: String? = null): String {
         val name = string.trim().removeSurrounding("$(", ")")
         return variableValues[name]
-                ?: key?.let { dictionary[key]?.toStringLiteral() }
+                ?: key?.let { dictionary.map[key]?.toStringLiteral() }
```
joelrosario: The map property of the dictionary should be wrapped. Nothing in the dictionary should be directly referred to outside. You could for example turn this into: 

`dictionary.lookup(key).toStringLiteral()` with the method signature of `lookup` being `fun lookup(key: String): Value`

core/src/test/kotlin/io/specmatic/core/DictionaryTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import io.specmatic.core.pattern.parsedJSONObject
+import io.specmatic.core.value.JSONObjectValue
+import org.assertj.core.api.Assertions.assertThat
+
+import org.junit.jupiter.api.Test
+
+// NOTE: Concrete values are actual values, not placeholders like (string) or (number).
+// For example, "ATYGHA" is a concrete value, not a placeholder.
+class DictionaryTest {
```
joelrosario, line 11: We need a test somewhere of how generation will work when a dictionary is involved.

core/src/main/kotlin/io/specmatic/core/Dictionary.kt
```diff
@@ -0,0 +1,78 @@
...
+                substituteDictionaryValues(value, paths, forceSubstitution)
+            }
+            is JSONArrayValue -> {
+                substituteDictionaryValues(value, paths, forceSubstitution)
+            }
+            else -> value
+        }
+    }
+
+    fun substituteDictionaryValues(httpResponse: HttpResponse, forceSubstitution: Boolean = false): HttpResponse {
```
joelrosario: We need to make this functionality generally available without tying it to HTTP requests and responses. Ideally these methods should be in their respective classes.

joelrosario (changes requested):

joelrosario (approved):

PR #1324, opened 2024-09-26
Add response interceptor hook in HttpStub

yogeshnikam671:
…ver http can modify the response before it gets passed on [repsonse-interceptor for HttpStub]

**What**:
Add a response interceptor hook in HttpStub

**Why**:
Other protocols can use this hook in case they want to modify the response before it gets passed over to the client.

**How**:
By adding a response interceptor.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -538,16 +538,11 @@ data class Feature(
...
-            val tag = if(description.isNotBlank())
-                " [${description}]"
-            else
-                ""
-            HasValue(result.value.copy(descriptionFromPlugin = "$descriptionFromPlugin${result.value.apiDescription}$tag"))
+            val apiDescription = result.value.descriptionFromPlugin ?: result.value.apiDescription
+            val tag = result.valueDetails.singleLineDescription().let {
+                if (it.isNotBlank()) " [$it]" else ""
+            }
+            HasValue(result.value.copy(descriptionFromPlugin = "$apiDescription$tag"))
```
joelrosario, line 562: This is not a good idea. Use .ifHasValue (here you seem to need the result).. ReturnValues may also have failures or Exceptions, and calling .value will result in an exception thrown in those cases.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/Feature.kt
yogeshnikam671: Hey Joel, I am already using the .ifHasValue on line 540. I think that should suffice. But I may be wrong. Please let me know if it does not suffice.

joelrosario (approved):

PR #1337, opened 2024-10-04
Generate negative tests for arrays

yogeshnikam671:
**What**:
The changes enable generation of a of negative patterns for arrays. For example, when a request contains an array, such as:

```yaml
components:
  schemas:
    Person:
      type: array
      items:
        type: string
```

the negative tests request generated for this schema will be: `[null]`, `[number]`, `[boolean]`.

**Why**:
Earlier the only negative pattern that was getting generated from arrays was null. In the above example, the only negative test generated would have been `[null]`.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


core/src/main/kotlin/io/specmatic/core/pattern/Row.kt
```diff
@@ -95,6 +95,9 @@ data class Row(
         if(requestBodyJSONExample == null)
             return this
 
+        if(requestBodyJSONExample.jsonObject is JSONArrayValue)
```
yogeshnikam671: @joelrosario , require your help here. For now we have introduced this condition to make a test pass. We are not much sure if this is right. We'll discuss.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -127,7 +127,22 @@ data class ListPattern(override val pattern: Pattern, override val typeAlias: St
         }
     }
 
-    override fun negativeBasedOn(row: Row, resolver: Resolver, config: NegativePatternConfiguration): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(NullPattern))
+    override fun negativeBasedOn(row: Row, resolver: Resolver, config: NegativePatternConfiguration): Sequence<ReturnValue<Pattern>> {
+        return pattern.negativeBasedOn(row, resolver, config)
+            .filter {
+                it.withDefault(true) {
```
joelrosario: Cleaned up how this works, please take a look

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -127,7 +127,22 @@ data class ListPattern(override val pattern: Pattern, override val typeAlias: St
...
+                    it !is NullPattern
+                }
+            }
+            .flatMap { negativePatternValue ->
+                listOf(
+                    negativePatternValue,
```
joelrosario: Has this logic been discussed with anyone?

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -127,7 +127,14 @@ data class ListPattern(override val pattern: Pattern, override val typeAlias: St
         }
     }
 
-    override fun negativeBasedOn(row: Row, resolver: Resolver, config: NegativePatternConfiguration): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(NullPattern))
+    override fun negativeBasedOn(row: Row, resolver: Resolver, config: NegativePatternConfiguration): Sequence<ReturnValue<Pattern>> {
```
joelrosario: We may need breadCrumbs added here. Should also check that the negative patterns look good in the sample project.

joelrosario (approved):

joelrosario (approved):

PR #1344, opened 2024-10-08
Generate negative scenarios where mandatory headers are removed

yogeshnikam671:
Add support for generation of negative test scenarios where mandatory headers are removed from the request.

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
+++ b/core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
@@ -233,20 +233,12 @@ data class HttpHeadersPattern(
     private fun patternsWithNoRequiredHeaders(
         patternMap: Map<String, Pattern>
     ): Sequence<ReturnValue<HttpHeadersPattern>> = sequence {
-        patternMap.forEach { (keyToOmit, _) ->
-            if(keyToOmit.endsWith("?").not()) {
-                yield(
-                    HasValue(
-                        HttpHeadersPattern(
-                            patternMap.filterKeys {
-                                key -> key != keyToOmit
-                            }.mapKeys { withoutOptionality(it.key) },
-                            contentType = contentType
-                        ),
-                        "mandatory header not sent"
-                    ).breadCrumb(keyToOmit)
-                )
-            }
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
```diff
@@ -214,16 +214,39 @@ data class HttpHeadersPattern(
...
             patternMapR.ifValue { patternMap ->
                 HttpHeadersPattern(
                     patternMap.mapKeys { withoutOptionality(it.key) },
                     contentType = contentType
                 )
             }
+        }.plus(patternsWithNoRequiredHeaders(pattern))
+    }
+
+    private fun patternsWithNoRequiredHeaders(
```
joelrosario, line 233: This function seems identical to `patternsWithNoRequiredQueryParams`. Could they not just be the same function?

core/src/test/kotlin/io/specmatic/conversions/OpenApiSpecificationTest.kt
```diff
@@ -8864,6 +8864,58 @@ paths:
         }
     }
 
+    @Nested
+    inner class NegativeScenariosForHeaders {
+        @Test
+        fun `should generate the negative scenarios where the mandatory headers are missing`() {
```
joelrosario, line 8870: We need one more test here showing that this functionality works when there is an example (given an example, we'd still generate a negative scenario for that example where the mandatory header is missing).

joelrosario (changes requested):

core/src/test/kotlin/io/specmatic/conversions/OpenApiSpecificationTest.kt
yogeshnikam671: Addressed.

core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
yogeshnikam671: Addressed.

joelrosario (approved):

PR #1353, opened 2024-10-11
Stop auto-generation of mandatory query params and headers missing in example

yogeshnikam671:
**What**:
Disable auto-generation of mandatory query param/header which is intentionally not specified in an example to emulate a negative scenario

**Why**:
If a user wants to specify an example where a mandatory query param is missing, currently specmatic auto-generates that mandatory query param even though it is not present in the example.
To fix this behaviour, this PR has been raised.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
+++ b/core/src/main/kotlin/io/specmatic/core/HttpHeadersPattern.kt
@@ -296,9 +296,12 @@ data class HttpHeadersPattern(
         return matches(this.pattern, row, resolver, "header")
     }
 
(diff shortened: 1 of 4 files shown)


joelrosario (changes requested):
While Specmatic allows examples where mandatory query params or headers are missing for 400 status responses, there is no reason to permit an example (one that has omitted mandatory query params or headers) when it is not an example of a bad request.

This logic should therefore not apply to non 4xx statuses. It should only apply specifically to 4xx statuses.

joelrosario (approved):

yogeshnikam671:
Hey @joelrosario, I've addressed the point raised by you. Please let me know your feedback on the same if any. Thanks.

joelrosario (approved):

PR #1364, opened 2024-10-19
Metadata based filtering for test

yogeshnikam671:
**What**:
Add ability to filter tests based on the metadata of a certain test.
Add --filter and --filterNot argument in TestCommand to use these filters.
The filters follow a specific syntax which can be understood using `specmatic  test --help` command.

**Why**:
The existing --filterName and --filterNotName were serving the purpose of filtering tests to an extent but potentially that algorithm is not scalable when it comes to evolving the filtering mechanism and making it more scalable.
Hence, this metadata based filtering is introduced. This is a first cut of it.

**Checklist**:

- [ ] Documentation  (Pending, will do it in parallel)
- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/application/src/main/kotlin/application/TestCommand.kt
+++ b/application/src/main/kotlin/application/TestCommand.kt
@@ -93,7 +93,7 @@ Filter tests matching the specified filtering criteria
 You can filter tests based on the following keys:
 - `METHOD`: HTTP methods (e.g., GET, POST)
 - `PATH`: Request paths (e.g., /users, /product)
-- `STATUS-CODE`: HTTP response status codes (e.g., 200, 400)
+- `STATUS`: HTTP response status codes (e.g., 200, 400)
 - `HEADERS`: Request headers (e.g., Accept, X-Request-ID)
 - `QUERY-PARAM`: Query parameters (e.g., status, productId)
 - `EXAMPLE-NAME`: Example name (e.g., create-product, active-status)
@@ -119,7 +119,7 @@ Filter tests not matching the specified criteria
 
 This option supports the same filtering keys and syntax as the --filter option.
 For example:
---filterNot="STATUS-CODE=400" --filterNot="METHOD=PATCH,PUT"
+--filterNot="STATUS=400" --filterNot="METHOD=PATCH,PUT"
            """
         ],
         required = false
--- a/application/src/main/kotlin/application/TestCommand.kt
+++ b/application/src/main/kotlin/application/TestCommand.kt
@@ -17,7 +17,9 @@ import io.specmatic.core.utilities.xmlToString
 import io.specmatic.test.SpecmaticJUnitSupport
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.CONTRACT_PATHS
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.ENV_NAME
+import io.specmatic.test.SpecmaticJUnitSupport.Companion.FILTER
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.FILTER_NAME_PROPERTY
+import io.specmatic.test.SpecmaticJUnitSupport.Companion.FILTER_NOT
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.FILTER_NOT_NAME_PROPERTY
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.HOST
 import io.specmatic.test.SpecmaticJUnitSupport.Companion.INLINE_SUGGESTIONS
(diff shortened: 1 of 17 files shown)


core/src/main/kotlin/io/specmatic/core/HttpRequestPattern.kt
```diff
@@ -30,6 +30,10 @@ data class HttpRequestPattern(
     val multiPartFormDataPattern: List<MultiPartFormDataPattern> = emptyList(),
     val securitySchemes: List<OpenAPISecurityScheme> = listOf(NoSecurityScheme())
 ) {
+    fun getHeaderKeys() = headersPattern.pattern.keys
```
joelrosario: Instead of digging into headers for data, wrap this in something like `headersPattern.headerNames`

core/src/main/kotlin/io/specmatic/core/HttpRequestPattern.kt
```diff
@@ -30,6 +30,10 @@ data class HttpRequestPattern(
...
+
+    fun getQueryParamKeys() = httpQueryParamPattern.queryPatterns.keys
```
joelrosario: Similar to the previous comment, httpQueryParamPattern.queryParamNames

core/src/main/kotlin/io/specmatic/core/filters/ScenarioFilterTags.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import io.specmatic.core.HEADERS_BREADCRUMB
+import io.specmatic.core.METHOD_BREAD_CRUMB
+import io.specmatic.core.PATH_BREAD_CRUMB
+import io.specmatic.core.QUERY_PARAMS_BREADCRUMB
+
+enum class ScenarioFilterTags(val key: String) {
+    METHOD(METHOD_BREAD_CRUMB),
+    PATH(PATH_BREAD_CRUMB),
+    STATUS_CODE("STATUS-CODE"),
```
joelrosario: Can you confirm that we do not already have a word for this, by checking in the status matching code HttpResponsePattern.

application/src/main/kotlin/application/TestCommand.kt
```diff
@@ -82,6 +84,48 @@ class TestCommand : Callable<Unit> {
...
+For example:
+--filter="METHOD=GET,POST" --filter="PATH=/users"
+           """
+        ],
+        required = false
+    )
+    var filter: List<String> = emptyList()
+
+    @Option(
+        names= ["--filterNot"],
```
joelrosario: this should be `--filter-not`, in keeping with `--filter-name` and `--filter-not-name`

core/src/main/kotlin/io/specmatic/core/utilities/Utilities.kt
```diff
@@ -334,7 +334,9 @@ fun saveJsonFile(jsonString: String, path: String, fileName: String) {
 }
 
 fun readEnvVarOrProperty(envVarName: String, propertyName: String): String? {
-    return System.getenv(envVarName) ?: System.getProperty(propertyName)
```
joelrosario: What's the reason for this change?

core/src/main/kotlin/io/specmatic/test/ContractTest.kt
```diff
@@ -9,6 +9,7 @@ interface ResponseValidator {
 }
 
 interface ContractTest {
+    val scenario: Scenario
```
joelrosario: Seems like the reason to expose this is because we need the scenario metadata and testDescription elsewhere. How about just exposing those methods in ContractTest, so that no-one has to dig in and chain method calls? Then you could remove the `scenario` property.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/utilities/Utilities.kt
yogeshnikam671: On retrospect I don't need this change. I'll revert. I had added this because I thought if I set env var as empty string, it won't work properly unless I return null from this function. That was an incorrect assumption.

core/src/main/kotlin/io/specmatic/test/ContractTest.kt
yogeshnikam671: This is a very good idea. I'll do that.

core/src/main/kotlin/io/specmatic/core/utilities/Utilities.kt
yogeshnikam671: Addressed.

core/src/main/kotlin/io/specmatic/test/ContractTest.kt
yogeshnikam671: Addressed.

application/src/main/kotlin/application/TestCommand.kt
yogeshnikam671: Addressed.

core/src/main/kotlin/io/specmatic/core/filters/ScenarioFilterTags.kt
yogeshnikam671: The breadCrumb just says STATUS, shall I change it to STATUS-CODE in breadcrumb in HttpResponsePattern and then use it here?

core/src/main/kotlin/io/specmatic/core/HttpRequestPattern.kt
yogeshnikam671: Addressed.

core/src/main/kotlin/io/specmatic/core/HttpRequestPattern.kt
yogeshnikam671: Addressed.

core/src/main/kotlin/io/specmatic/core/filters/ScenarioFilterTags.kt
joelrosario: Let this be STATUS for now, but we should change it to STATUS-CODE later.

joelrosario (approved):

PR #1367, opened 2024-10-20
Implement multiple examples loading and generation in the Interactive Server.

StarKhan6368:
**What**: Implement multiple examples loading and generation in the Interactive Server.

Load multiple examples and generate more examples in interactive examples UI.

**Why**: 
- At present, we are limited to loading only a single existing example, regardless of the number available. As a result, the user can view just that one example and is restricted to generating only one additional example if no examples are present.
- This PR allows for the loading of multiple existing examples and provides the user with the capability to generate additional examples if necessary.

- Adding `Generate More` button when there are already existing examples
- Show each example as a separate row under each operation

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
+++ b/core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
@@ -1,9 +1,11 @@
 package io.specmatic.core.examples.server
 
+import io.specmatic.conversions.ExampleFromFile
 import io.specmatic.conversions.convertPathParameterStyle
 import io.specmatic.core.Feature
+import io.specmatic.core.Scenario
 import io.specmatic.core.examples.server.ExamplesInteractiveServer.Companion.getExamplesFromDir
-import io.specmatic.core.examples.server.ExamplesInteractiveServer.Companion.getExistingExampleFile
+import io.specmatic.core.examples.server.ExamplesInteractiveServer.Companion.getExistingExampleFiles
 import org.thymeleaf.TemplateEngine
 import org.thymeleaf.context.Context
 import org.thymeleaf.templatemode.TemplateMode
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
```diff
@@ -14,56 +16,81 @@ class ExamplesView {
     companion object {
         fun getEndpoints(feature: Feature, examplesDir: File): List<Endpoint> {
             val examples = examplesDir.getExamplesFromDir()
-            return feature.scenarios.map {
-                val example = getExistingExampleFile(it, examples)
+            val scenarioExPairList = getScenarioExamplesPairs(feature.scenarios, examples)
```
harikrishnan83: Please expand `Ex` to `Example`, will be easier to read

core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
```diff
@@ -14,56 +16,81 @@ class ExamplesView {
...
+
+            return scenarioExPairList.map { (scenario, ex) ->
```
harikrishnan83: same as above, expand `ex` to `example`

core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
```diff
@@ -14,56 +16,81 @@ class ExamplesView {
...
-                    responseStatus = it.httpResponsePattern.status,
-                    contentType = it.httpRequestPattern.headersPattern.contentType,
-                    exampleFile = example?.first,
-                    exampleMismatchReason = example?.second
+                    path = convertPathParameterStyle(scenario.path),
+                    rawPath = scenario.path,
+                    method = scenario.method,
+                    responseStatus = scenario.httpResponsePattern.status,
+                    contentType = scenario.httpRequestPattern.headersPattern.contentType,
+                    exampleFile = ex?.first,
```
harikrishnan83: Can we further explode `ex` to `exampleFile` and `exampleMismatchReason`? so that we avoid using `first` and `second`

core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesView.kt
```diff
@@ -14,56 +16,81 @@ class ExamplesView {
...
+                    exampleMismatchReason = ex?.second
                 )
-            }.filterEndpoints().sortEndpoints()
```
harikrishnan83, line 28: any reason we stopped sorting endpoints? was this intentional?

harikrishnan83 (approved):
Looks good to me

PR #1368, opened 2024-10-20
Loading config file from classpath Also in case of env property or classpath checking if the file exists before returning

nashjain:
**What**:

If the config file is available as part of the classpath then use it instead of always expecting it to be at the top level of the project.

**Why**:

Some developers prefer to keep specmatic.yaml config file inside their src/test/resources folder.

**Checklist**:

- [ ] Tests
- [ ] Sonar Quality Gate


harikrishnan83 (approved):
Looks good to me

PR #1407, opened 2024-10-31
New Examples Transform Command

StarKhan6368:
**What**: Examples Transform Command

**Why**: The examples transform command enables you to transform your examples based on the new OpenAPI specification. This specification is generated by applying the provided overlay to the original specification.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/application/src/main/kotlin/application/ExamplesCommand.kt
+++ b/application/src/main/kotlin/application/ExamplesCommand.kt
@@ -557,7 +557,12 @@ For example:
         @Option(names = ["--only-mandatory-keys-in-payload"], description = ["Transform existing examples so that they contain only mandatory keys in payload, default is true"], required = false)
         var allowOnlyMandatoryKeysInPayload: Boolean = true
 
+        @Option(names = ["--debug"], description = ["Debug Logs"])
+        var verbose: Boolean = false
+
         override fun call() {
+            configureLogger(verbose)
+
             if(allowOnlyMandatoryKeysInPayload) {
                 ExamplesInteractiveServer.transformExistingExamples(
                     contractFile,
--- a/application/src/main/kotlin/application/ExamplesCommand.kt
+++ b/application/src/main/kotlin/application/ExamplesCommand.kt
@@ -26,7 +26,11 @@ private const val FAILURE_EXIT_CODE = 1
     name = "examples",
     mixinStandardHelpOptions = true,
     description = ["Generate externalised JSON example files with API requests and responses"],
-    subcommands = [ExamplesCommand.Validate::class, ExamplesCommand.Interactive::class]
+    subcommands = [
+        ExamplesCommand.Validate::class,
+        ExamplesCommand.Interactive::class,
+        ExamplesCommand.Transform::class
+    ]
 )
 class ExamplesCommand : Callable<Int> {
     @Option(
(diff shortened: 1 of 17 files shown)


core/src/main/kotlin/io/specmatic/core/pattern/JSONObjectPattern.kt
```diff
@@ -27,6 +27,19 @@ data class JSONObjectPattern(
     val minProperties: Int? = null,
     val maxProperties: Int? = null
 ) : Pattern {
+
+    override fun eliminateOptionalKey(value: Value, resolver: Resolver): Value {
+        if (value !is JSONObjectValue) return value
+
+        val filteredPattern = this.pattern.filterKeys { !it.endsWith("?") }
```
joelrosario: filteredPattern and filteredJsonObject variable names should be renamed for clarity. Perhaps mandatoryObjectPatternMap and mandatoryObjectMap.

core/src/main/kotlin/io/specmatic/core/pattern/JSONArrayPattern.kt
```diff
@@ -13,6 +13,11 @@ import io.specmatic.core.value.Value
 import java.util.*
 
 data class JSONArrayPattern(override val pattern: List<Pattern> = emptyList(), override val typeAlias: String? = null) : Pattern, SequenceType {
+
+    override fun eliminateOptionalKey(value: Value, resolver: Resolver): Value {
+        return JSONArrayValue(emptyList())
```
joelrosario: Either we should fully implement, or not at all. Given this is a legacy pattern, I suggest not implementing for now.

core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
```diff
@@ -86,6 +86,21 @@ data class JSONObjectValue(val jsonObject: Map<String, Value> = emptyMap()) : Va
 
     fun findFirstChildByName(name: String): Value? =
         jsonObject[name]
+
+    fun objectWithMandatoryKeys(
```
joelrosario: Do we still need this method? I believe it's not being called from anywhere.

joelrosario (changes requested):

application/src/main/kotlin/application/ExamplesCommand.kt
```diff
@@ -534,6 +538,36 @@ For example:
...
+        lateinit var contractFile: File
+
+        @Option(names = ["--overlay-file"], description = ["Overlay file path"], required = false)
+        val overlayFile: File? = null
+
+        @Option(names = ["--examples-dir"], description = ["Directory where existing examples reside"], required = true)
+        lateinit var examplesDir: File
+
+        @Option(names = ["--only-mandatory-keys-in-payload"], description = ["Transform existing examples so that they contain only mandatory keys in payload"], required = false)
+        var allowOnlyMandatoryKeysInPayload: Boolean = false
```
joelrosario: this should be implicitly set to true if given, we should not have to say --only-mandatory-keys-in-payload=true

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesInteractiveServer.kt
```diff
@@ -724,6 +725,52 @@ class ExamplesInteractiveServer(
         private fun Map<String, String>.toValueMap(): Map<String, Value> {
             return this.mapValues { StringValue(it.value) }
         }
+
+        fun transformExistingExamples(contractFile: File, overlayFile: File?, examplesDir: File) {
+            val feature = parseContractFileToFeature(contractPath = contractFile.absolutePath, overlayContent = overlayFile?.readText().orEmpty())
+            val examples = examplesDir.getExamplesFromDir()
+
+            examples.forEach { example ->
+                consoleLog("\nTransforming ${example.file.nameWithoutExtension}")
```
joelrosario: Put a verbose mode in, so we can switch the logging on if needed. There are too many logs if it's on by default.

core/src/main/kotlin/io/specmatic/core/pattern/JSONObjectPattern.kt
StarKhan6368: Thanks, I've fixed that

core/src/main/kotlin/io/specmatic/core/pattern/JSONArrayPattern.kt
StarKhan6368: Thanks, removed the implementation

core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
StarKhan6368: Agree, i've dropped the method now

application/src/main/kotlin/application/ExamplesCommand.kt
StarKhan6368: yeah makes sense, I've addressed this, also added a few simple tests for now.

joelrosario (approved):

PR #1426, opened 2024-11-14
Pattern Validation - Max Length Issue Fix

samyakOO7:
Pattern validation was taking priority in response schema for stub and request schema for test over max length parameter due to random() method generating value more than default

Fixed the use case putting a constraint of max length on random() method so that it never generates value for than max length keeping the current logic intact

Test Spec File - [validate-regex-spec.yaml.zip](https://github.com/user-attachments/files/17747959/validate-regex-spec.yaml.zip)

Backward Compatibility Spec (v2.0.36 release check) - 
[validate-regex-spec.yaml.zip](https://github.com/user-attachments/files/17797468/validate-regex-spec.yaml.zip)

Logic used -

>Default length - 5 (remains intact)

> If minimum and maximum is both supplied, pattern length would be in range of (min to max) 

> If minimum length is not supplied in spec -> and maximum is less than default (5), pattern would be of length (1 i.e. generex min value, max)
Say max = 4 is only supplied 
pattern length would be in range (1-4)

> If minimum length is not supplied in spec -> and maximum is greater than default (5), pattern would be of length (5 (default, max)
Say max = 15 is only supplied 
pattern length would be in range (5-15)

> if max is not supplied and min value exist we will assume to take pattern length from (min , generex max value)

**Test for maximum length conditions :**
Request : `{minLength : 2, maxLength: 11}`
Response : `{minLength : 2, maxLength: 6}`

Before Stub : 
```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "12000000" }' http://0.0.0.0:49272/validate
{
    "number": "873.13863"
}
```               
Generated value : 9 digit

Before Test:
Details:
```shell
Scenario: POST /validate -> 200 has FAILED
In scenario "Validate the input number. Response: Input is valid"
API: POST /validate -> 200

  In scenario "Validate the input number. Response: Input is valid"
     API: POST /validate -> 200
   
       REQUEST.BODY.number Contract expected string with maxLength 2 but request contained "2.8837947"
```
          
After Issue Fixed Stub:

```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "12000000" }' http://0.0.0.0:9000/validate
{
    "number": "8.9162"
}
```           

After Issue Fixed Test:

```shell
Request:
POST /validate
Accept-Charset: UTF-8
Accept: */*
Content-Type: application/json

{
    "number": "3."
}

Response:
200 OK
Vary: Origin
X-Specmatic-Result: success
X-Specmatic-Type: random
Content-Length: 26
Content-Type: application/json
Connection: keep-alive

{
    "number": "3618.9"
}
```

**Test for Minimum Length  conditions:**
Request : `{minLength : 2, maxLength: 11}`
Response : `{minLength : 2, maxLength: 6}`

Stub:
Before Fix (The output generated was defaulting to 5 as minimum for num<5 (breaching min validation):
```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:54811/validate
{
    "number": "2.866"
}
 curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:54811/validate
{
    "number": "235.4"
}
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:54811/validate
{
    "number": "8.150"
}
```   

After Fix (Min validation is respected):  
```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "2."
}
 curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "5.1"
}
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:54811/validate
{
    "number": "8.150"
}
 curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "39"
}
```   

Test :
Before Fix: 
```shell
Request:
POST /validate
Accept-Charset: UTF-8
Accept: */*
Content-Type: application/json

{
    "number": "6.7293812038"
} 

Response:
400 Bad Request
Vary: Origin
X-Specmatic-Result: failure
Content-Length: 411
Content-Type: text/plain
Connection: keep-alive

In scenario "Validate the input number. Response: Input is valid"
API: POST /validate -> 200

  >> REQUEST.BODY.number
  
     Contract expected string with maxLength 2 but request contained "6.7293812038"
```
After Fix : 
```shell
Request:
POST /validate
Accept-Charset: UTF-8
Accept: */*
Content-Type: application/json

{
    "number": "9."
}
Response:
200 OK
Vary: Origin
X-Specmatic-Result: success
X-Specmatic-Type: random
Content-Length: 23
Content-Type: application/json
Connection: keep-alive

{
    "number": "407"
}
```
Some edge cases taken care of defaults -

 Request {  maxLength = 4 }
Response {  maxLength = 4 }  
min  = null and max (4) < default (5) [pattern (min ‎ =  Generex(min val = 1), max = 4)]

Stub- 
```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "7"
} curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "3.47"
}
```   
Test - 
```shell
Request:
POST /validate
Accept-Charset: UTF-8
Accept: */*
Content-Type: application/json

{
    "number": "65.0"
}
Response:
200 OK
Vary: Origin
X-Specmatic-Result: success
X-Specmatic-Type: random
Content-Length: 23
Content-Type: application/json
Connection: keep-alive

{
    "number": "4.72”
}
```
Request {  maxLength = 15 }
Response {  maxLength = 15 }
min length = null and max length > default (5) [pattern (min ‎ =  default(5), max = 15)]

Stub-
```shell
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "8.915"
} curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "9.609"
}
curl -X POST -H "Content-Type: application/json" -d '{ "number": "1" }' http://0.0.0.0:9000/validate
{
    "number": "7.09024600"
}
```   

Test - 
``` shell
Request:
POST /validate
Accept-Charset: UTF-8
Accept: */*
Content-Type: application/json

{
    "number": "4.22214"
} Response:
200 OK
Vary: Origin
X-Specmatic-Result: success
X-Specmatic-Type: random
Content-Length: 25
Content-Type: application/json
Connection: keep-alive

{
    "number": "2.648"
} 
```
Clear outputs respecting validation schema are found with multiple test of min and max length

Thanks


core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -67,20 +67,28 @@ data class StringPattern (
...
-    
+
     override fun generate(resolver: Resolver): Value {
         val defaultExample: Value? = resolver.resolveExample(example, this)
-
         if (regex != null) {
-            if(defaultExample == null)
-                return StringValue(Generex(regex.removePrefix("^").removeSuffix("$")).random(randomStringLength))
+            if(defaultExample == null) {
+                if (maxLength != null)
```
harikrishnan83: Is there a possibility that the regex pattern can generate a String that is less than `minLength`, should we handle that also?

core/src/test/kotlin/io/specmatic/core/pattern/StringPatternTest.kt
```diff
@@ -240,4 +250,12 @@ internal class StringPatternTest {
     fun `string pattern encompasses email`() {
         assertThat(StringPattern().encompasses(EmailPattern(), Resolver(), Resolver())).isInstanceOf(Result.Success::class.java)
     }
+
+    @Test
+    fun `should fail to generate string when maxLength is less than minLength`() {
```
harikrishnan83, line 263: good that you have covered this scenario with a test

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -67,20 +67,28 @@ data class StringPattern (
...
+                    return StringValue(
+                        Generex(regex.removePrefix("^").removeSuffix("$")).random(
```
harikrishnan83: any reason we have to remove the `^` and `$`?

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -67,20 +67,28 @@ data class StringPattern (
...
+                            randomStringLength,
```
harikrishnan83: if `randomStringLength` is already based on `minLength` and `maxLength` why are we looking at `maxLength` again here?

harikrishnan83 (approved):
Looks good to me

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -67,36 +67,27 @@ data class StringPattern (
...
-                throw ContractException("Schema example ${defaultExample.toStringLiteral()} is not a string")
+    override fun generate(resolver: Resolver): Value {
+        val defaultExample = resolver.resolveExample(example, this)
 
-            return defaultExample
+        defaultExample?.let {
+            if (matches(it, resolver).isSuccess()) {
+                return it
+            }
+            throw ContractException("Schema example ${it.toStringLiteral()} does not match pattern $regex")
```
joelrosario: In this form, this condition is no more about regex matching.

Ideally, the resolver should be given a new MismatchMessages indicating that we are checking the schema example.

The `matches` method returns a result, and we could call result.throwOnFailure() instead of throwing this exception

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -67,36 +67,27 @@ data class StringPattern (
         return JSONArrayValue(valueList)
     }
 
-    private val randomStringLength: Int =
+    private val patternMinLength: Int =
         when {
-            minLength != null && 5 < minLength -> minLength
-            maxLength != null && 5 > maxLength -> maxLength
+            minLength != null && minLength > 0 -> minLength
+            maxLength != null && maxLength < 5 -> 1
```
joelrosario: This worked a little differently earlier. If maxLength < 4, use it, else use upto 5 chars. What's the reason we changed this?

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
```diff
@@ -157,6 +148,11 @@ data class StringPattern (
 
     override val pattern: Any = "(string)"
     override fun toString(): String = pattern.toString()
+
+    private fun generateFromRegex(regexWithoutCaretAndDollar: String, minLength: Int, maxLength: Int?): String =
+        maxLength?.let {
+            Generex(regexWithoutCaretAndDollar).random(minLength, it)
```
joelrosario: Regexes supported by Generex may also have a length, e.g. a{10}. While we are at it, do we want to handle this in case it's greater than a maxLength or less than minLength?

core/src/test/kotlin/io/specmatic/core/pattern/StringPatternTest.kt
```diff
@@ -72,28 +72,47 @@ internal class StringPatternTest {
...
+                Arguments.of(null, 1, 10, 1),
+            )
+        }
     }
 
     @ParameterizedTest
-    @MethodSource("lengthTestValues")
-    fun `generate string value of appropriate length matching minLength and maxLength parameters`(min: Int?, max: Int?, length: Int) {
+    @MethodSource("minLengthMaxLengthAndExpectedLength")
+    fun `generate string value as per minLength and maxLength`(min: Int?, max: Int?, expectedLengthOfGeneratedValue: Int) {
```
joelrosario: Use CsvSource instead of MethodSource, to keep the test data local to the test.

core/src/test/kotlin/io/specmatic/core/pattern/StringPatternTest.kt
```diff
@@ -72,28 +72,47 @@ internal class StringPatternTest {
...
+        val generatedLength = result.string.length
 
-        assertThat(result.string.length).isEqualTo(length)
+        assertThat(generatedLength).isGreaterThanOrEqualTo(expectedLengthOfGeneratedValue)
+        max?.let { assertThat(generatedLength).isLessThanOrEqualTo(it) }
+    }
+
+    @ParameterizedTest
+    @MethodSource("regexMinLengthAndMaxLengthAndExpectedLength")
+    fun `generate string value as per regex in conjunction with minLength and maxLength`(regex: String?, min: Int?, max: Int?, expectedLength: Int) {
```
joelrosario: Use CsvSource instead of MethodSource, to keep the test data local to the test.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
samyakOO7: Logic used -

Default length - 5 (remains intact)
If minimum and maximum is both supplied, pattern length would be in range of (min to max)
If minimum length is not supplied in spec -> and maximum is less than default (5), pattern would be of length (1 i.e. generex min value, max)
Say max = 4 is only supplied
pattern length would be in range (1-4)
If minimum length is not supplied in spec -> and maximum is greater than default (5), pattern would be of length (5 (default, max)
Say max = 15 is only supplied
pattern length would be in range (5-15)
if max is not supplied and min value exist we will assume to take pattern length from (min , generex max value)

samyakOO7: We revamped this logic a bit as per the one given above

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
samyakOO7: We are putting the constraints to generate regex from min to max length, previously we only gave it the min length and we assumed it to be normal length, which caused it to overflow most of the time, out of max range

The logic written before was a bit tricky to understand but was failing lot of cases

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
samyakOO7: As these characters are not addressed by generex we are omitting it

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
samyakOO7: Before it was only based on minLength

core/src/main/kotlin/io/specmatic/core/pattern/StringPattern.kt
samyakOO7: Resolved

harikrishnan83 (approved):

joelrosario (approved):

PR #1428, opened 2024-11-14
Stateful stub

yogeshnikam671:


core/src/main/kotlin/io/specmatic/core/ResponseBuilder.kt
```diff
@@ -1,10 +1,18 @@
...
+    val responseBodyPattern = scenario?.resolver?.withCyclePrevention(
+        scenario.httpResponsePattern.body
+    ) {
+        resolvedHop(scenario.httpResponsePattern.body, it)
+    }
+    val resolver = scenario?.resolver
+
     fun build(requestContext: RequestContext): HttpResponse {
-        return scenario.generateHttpResponse(serverState, requestContext)
+        return scenario!!.generateHttpResponse(serverState, requestContext)
```
joelrosario: Why is this not null-checked?

core/src/main/kotlin/io/specmatic/core/Scenario.kt
```diff
@@ -160,10 +160,13 @@ data class Scenario(
     fun matchesStub(
         httpRequest: HttpRequest,
         serverState: Map<String, Value>,
-        mismatchMessages: MismatchMessages = DefaultMismatchMessages
+        mismatchMessages: MismatchMessages = DefaultMismatchMessages,
+        unexpectedKeyCheck: UnexpectedKeyCheck = ValidateUnexpectedKeys
```
joelrosario: let this be null by default, and don't add any unexpectedKeyCheck to the resolver if it is null.

joelrosario (approved):

PR #1459, opened 2024-11-29
Fixed status code 1 issue with backward compatibility

samyakOO7:
Fixed status code 1 issue while running backward compatibility check when no spec files change is found

### DIFF
--- a/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
+++ b/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
@@ -92,7 +92,7 @@ abstract class BackwardCompatibilityCheckBaseCommand : Callable<Unit> {
         ).filter {
             File(it).exists() && File(it).isValidSpec()
         }.toSet().also {
-            it.takeIf { it.isEmpty() }?.run {
+            if (it.isEmpty()) {
                 logger.log("$newLine No specs were changed, skipping the check.$newLine")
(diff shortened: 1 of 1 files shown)


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -92,7 +92,10 @@ abstract class BackwardCompatibilityCheckBaseCommand : Callable<Unit> {
         ).filter {
             File(it).exists() && File(it).isValidSpec()
         }.toSet().also {
-            if(it.isEmpty()) exitWithMessage("No specs were changed, skipping the check.")
+            it.takeIf { it.isEmpty() }?.run {
```
harikrishnan83: Just curious, why do we have to switch to `takeIf`? Can we not make the `if(it.isEmpty())` condition body a block body and just put the logger line and exitProcess with 0?

harikrishnan83 (changes requested):
Thanks for the PR. Can you please look into my comment?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
samyakOO7: Yes, just from a functional programming perspective I added but it's unclear and extra checks not needed so rectified it

harikrishnan83 (approved):
Looks good to me

PR #1474, opened 2024-12-14
Code Editor (Specmatic Error - Line Highlighting)

samyakOO7:
Code Editor Specmatic Error Line Highlighting

The functionality implemented highlights error lines in a code editor using metadata and JSON paths derived from a Specmatic response.

**Functionality:**
1. Parse a JSON document and its associated metadata using jsonSourceMap to map JSON paths to their respective locations (lines and columns).
2. Iterate through the metadata, match each jsonPath to the corresponding location in the parsed pointers, and highlight the error line in the editor.
3. Handle cases where the jsonPath does not directly match any key by removing the trailing parts of the path incrementally until a match is found.
4.Provide a fallback to use value line information if key is not available.

**Key Features:**

1.Error Highlighting:
The error lines are marked in the editor using editor.markText, with a CSS class (CodeMirror-lint-mark-error) for styling.
A tooltip is added to the highlighted line to display the error description.

2.Fallback for Missing Keys:
If the key location is not found in the pointers, the function falls back to using the value location for highlighting.

3.Dynamic Path Matching:
If the jsonPath does not match any pointer, the function trims the path from the leaf node backward (e.g., /store/location/city → /store/location) until a match is found.

4.Gutter Marker:
Adds a marker in the editor's gutter to visually indicate errors alongside the line numbers.

**Optimization and Improvements**

1. Reusable Path Matching:
A utility function (findObjectByPath) matches paths dynamically, ensuring robustness even with incomplete paths.
2. Dynamic Path Trimming:
Implements a strategy to handle missing paths by iteratively trimming the path and retrying.
3. Support for Key and Value Locations:
Ensures that both key and value locations are checked for highlighting, covering all possible cases.


harikrishnan83 (changes requested):
Please add unit tests for `ExampleValidationErrorMessage` jsonpath extraction logic

PR #1476, opened 2024-12-15
Pass the 4xx error messages in message or msg key if it is found else fallback to default logic

yogeshnikam671:
**What**:
Pass the 4xx error messages in a key named "message" or "msg" in the response body for a virtual-service.
If the keys are not found, pass the message in the first key that of type string.

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate


nashjain (approved):
LGTM

PR #1492, opened 2024-12-22
Code Mirror 6 Integration

samyakOO7:
### PR Summary:

CM5 did not have sophisticated mechanisms for tooltips (So we had to override it using CSS). Moving CM6 for cleaner implementation of tooltip and improved UX. Also for more frequent updates.


core/src/main/kotlin/io/specmatic/core/examples/server/ExamplesInteractiveServer.kt
```diff
@@ -106,6 +107,9 @@ class ExamplesInteractiveServer(
             configureHealthCheckModule()
 
             routing {
+                static("/_specmatic/examples") {
```
harikrishnan83: Can we change this to `/_specmatic/examples/assets` so that we separate the URL that renders index from URL that serves assets

harikrishnan83 (approved):
Looks good to me

PR #1528, opened 2025-01-05
Interactive Examples Editor Error Highlighting Improvements

samyakOO7:
The following issues are fixed as a part of this PR-

- The error highlighting disappears on the line where we hover / is currently under focus. If it is the first line which has the error then focus by default is on that line, so no error highlights are seen.
- The statement “Expected value but it was” is displayed for empty Strings. It should be `Expected value but it was “”` 
- When the editor is updated, the line decorations should continue to remain on the errors
- Changed the line decoration to highlight only the part of a line that has errors so that users instinctively hover on chars to see the tooltip

### DIFF
--- a/core/src/main/resources/static/setup.css
+++ b/core/src/main/resources/static/setup.css
@@ -297,7 +297,7 @@ video {
 }
 
 .specmatic-editor-line-error {
-    background-color: rgba(255, 0, 0, 0.2) !important;
+   background-color: rgba(255, 0, 0,0.2);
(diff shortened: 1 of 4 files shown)


core/src/main/resources/static/setup.css
harikrishnan83, line 1: Please undo changes here, this file is not updated.

harikrishnan83 (changes requested):

harikrishnan83 (approved):

PR #1567, opened 2025-01-27
Make Examples Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the examples field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -120,7 +120,7 @@ data class SpecmaticConfig(
     val test: TestConfiguration? = TestConfiguration(),
     val stub: StubConfiguration = StubConfiguration(),
     val virtualService: VirtualServiceConfiguration = VirtualServiceConfiguration(),
-    val examples: List<String> = getStringValue(EXAMPLE_DIRECTORIES)?.split(",") ?: emptyList(),
+    private val examples: List<String> = getStringValue(EXAMPLE_DIRECTORIES)?.split(",") ?: emptyList(),
```
joelrosario: Shouldn't this code have moved into the getExamples function?

sukesh2000: Done, I've moved these to the `getExamples()` and made examples field nullable

core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -255,7 +260,7 @@ data class ResiliencyTestsConfig(
     }
 }
 
-class Auth(
+data class Auth(
```
joelrosario, line 305: Why did this become a data class here?

sukesh2000: It was a data class only, by mistake in  PR #1569, I made it a normal class. So I'm correcting it in this PR.

joelrosario (approved):

PR #1573, opened 2025-01-29
Make Pipeline Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the pipeline field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -218,6 +218,31 @@ data class SpecmaticConfig(
...
         return auth?.getBearerEnvironmentVariable()
     }
+
+    @JsonIgnore
+    fun getPipeline(): Pipeline? {
+        return pipeline
+    }
+
+    @JsonIgnore
+    fun getPipelineProvider(): PipelineProvider? {
```
joelrosario, line 228: Is this method needed?

joelrosario (approved):

PR #1577, opened 2025-01-30
Make Report Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the report field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -218,6 +224,16 @@ data class SpecmaticConfig(
...
         return auth?.getBearerEnvironmentVariable()
     }
+
+    @JsonIgnore
+    fun getReportFormatters(): List<ReportFormatter>? {
+        return report?.getFormatters()
+    }
+
+    @JsonIgnore
+    fun getReportTypes(): ReportTypes? {
```
joelrosario: Within ReportTypes, there is `OpenAPI`, which is a property of type `APICoverageConfiguration`. The properties and subproperties of `APICoverageConfiguration` should also be encapculated as methods.

sukesh2000: Done, encapsulated all of them

junit5-support/src/main/kotlin/io/specmatic/test/SpecmaticJUnitSupport.kt
```diff
@@ -101,7 +115,7 @@ open class SpecmaticJUnitSupport {
         }
 
         private fun getReportConfiguration(): ReportConfiguration {
-            return when (val reportConfiguration = specmaticConfig?.report) {
+            return when (val reportConfiguration = getReport(specmaticConfig)) {
```
joelrosario: getReport (which is in SpecmaticConfig.Companion) is meant to be used only within configuration classes. It looks like it's presence is being checked (first `null` branch of the `when`). You can use a method on SpecmaticConfig to answer that question instead of returning the object.

joelrosario (changes requested):

junit5-support/src/main/kotlin/io/specmatic/test/reports/OpenApiCoverageReportProcessor.kt
```diff
@@ -22,9 +23,12 @@ class OpenApiCoverageReportProcessor (private val openApiCoverageReportInput: Op
     }
 
     override fun process(specmaticConfig: SpecmaticConfig) {
-        val reportConfiguration = specmaticConfig.report!!
+        val reportConfiguration = getReport(specmaticConfig)!!
```
joelrosario: getReport should not be used directly as it's on the companion object. There are already methods on SpecmaticConfig for fetching the needed data and those should be used directly.

joelrosario (changes requested):

PR #1578, opened 2025-01-30
Make Security Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the security field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -1129,7 +1128,8 @@ class OpenApiSpecification(
     }
 
     private fun toSecurityScheme(schemeName: String, securityScheme: SecurityScheme): OpenAPISecurityScheme {
-        val securitySchemeConfiguration = securityConfiguration?.OpenAPI?.securitySchemes?.get(schemeName)
+        val securitySchemeConfiguration =
+            securityConfiguration?.getOpenAPISecuritySchemes()?.get(schemeName)
```
joelrosario: Let this read `securityConfiguration?.getOpenAPISecuritySchemes(schemeName)`.

Can you get rid of `getOpenAPISecuritySchemes`.

sukesh2000: Done, renamed the method to `getOpenAPISecurityScheme(scheme: String)`

joelrosario (approved):

PR #1581, opened 2025-01-31
Make Test Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the test field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -128,9 +129,10 @@ data class Feature(
     val strictMode: Boolean = false
 ): IFeature {
     fun enableGenerativeTesting(onlyPositive: Boolean = false): Feature {
+        val testConfiguration = getTestConfiguration(specmaticConfig)
         val updatedSpecmaticConfig = specmaticConfig.copy(
-            test = specmaticConfig.test?.copy(
-                resiliencyTests = specmaticConfig.test.resiliencyTests?.copy(
+            test = testConfiguration?.copy(
```
joelrosario: There should be a method in specmaticConfig to perform this operation.

sukesh2000: Addressed, moved this operation to method `copyResiliencyTestsConfig()` in `SpecmaticConfig`

core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -278,26 +299,46 @@ data class SpecmaticConfig(
...
-    val validateResponseValues: Boolean? = null,
-    val allowExtensibleSchema: Boolean? = null,
-    val timeoutInMilliseconds: Long? = getLongValue(SPECMATIC_TEST_TIMEOUT)
-)
+    private val resiliencyTests: ResiliencyTestsConfig? = null,
+    private val validateResponseValues: Boolean? = null,
+    private val allowExtensibleSchema: Boolean? = null,
+    private val timeoutInMilliseconds: Long? = null
+) {
+    fun getResiliencyTests(): ResiliencyTestsConfig {
```
joelrosario, line 398: In most cases, these getter methods on inner classes aren't necessary, as the objective of hiding the internal config structure from the rest of the application is achieved by the top-level method in SpecmaticConfig which returns a scalar.

However there are some cases in which the top-level method returns an object. In such cases, we would need these getter methods on the object being returned.

We may need to return to this across objects returning classes in a subsequent PR.

However, please apply this to all currently pending PRs.

joelrosario (changes requested):
Please address my comments.

joelrosario (approved):

PR #1582, opened 2025-01-31
Make Stub Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the stub field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -77,11 +77,27 @@ fun String.loadContract(): Feature {
 }
 
 data class StubConfiguration(
-    val generative: Boolean? = null,
-    val delayInMilliseconds: Long? = getLongValue(SPECMATIC_STUB_DELAY),
-    val dictionary: String? = getStringValue(SPECMATIC_STUB_DICTIONARY),
-    val includeMandatoryAndRequestedKeysInResponse: Boolean? = null
-)
+    private val generative: Boolean? = null,
+    private val delayInMilliseconds: Long? = getLongValue(SPECMATIC_STUB_DELAY),
```
joelrosario: This may result in fields with default values appearing in config post upgrade, where they were not previously there.

sukesh2000: Yes yes, I've resolved this and moved the default value substitution to the getters

joelrosario (changes requested):
There's a [comment on getters in child objects](https://github.com/znsio/specmatic/pull/1581#pullrequestreview-2595734998) in config on the test config. Please have a look and address them on this PR.

joelrosario (approved):

PR #1583, opened 2025-01-31
Make VirtualService Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the virtualService field in the SpecmaticConfig class from public to private.


joelrosario (changes requested):
Please add a test to this PR for loading the config (if not already present) and transforming it.

joelrosario (approved):

PR #1584, opened 2025-02-03
Make Workflow Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the workflow field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -572,7 +573,7 @@ data class Feature(
         Results(results.map { it.second }.filterIsInstance<Result.Failure>().toMutableList())
 
     fun generateContractTests(suggestions: List<Scenario>, fn: (Scenario, Row) -> Scenario = { s, _ -> s }): Sequence<ContractTest> {
-        val workflow = Workflow(specmaticConfig.workflow ?: WorkflowConfiguration())
+        val workflow = Workflow(getWorkflowConfiguration(specmaticConfig) ?: WorkflowConfiguration())
```
joelrosario: Companion object methods are meant to be used only in configuration classes for the purpose of versioned config construction. In this case, the map within the workflow object should be propagated instead of the workflow object.

joelrosario (changes requested):
Please see my comments.

joelrosario (approved):

PR #1586, opened 2025-02-03
Make AttributeSelectionPattern Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the attributeSelectionPattern field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -98,16 +98,24 @@ data class WorkflowConfiguration(
...
+    private val defaultFields: List<String>? = null,
     @field:JsonAlias("query_param_key")
-    val queryParamKey: String = readEnvVarOrProperty(
-        ATTRIBUTE_SELECTION_QUERY_PARAM_KEY,
-        ATTRIBUTE_SELECTION_QUERY_PARAM_KEY
-    ).orEmpty()
-)
+    private val queryParamKey: String? = null
+) {
+    fun getDefaultFields(): List<String> {
```
joelrosario: This behaviour should remain in this class.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -238,7 +239,7 @@ class OpenApiSpecification(
         val updatedScenarios = scenarioInfos.map {
             Scenario(it).copy(
                 dictionary = dictionary.plus(specmaticConfig.parsedDefaultPatternValues()),
-                attributeSelectionPattern = specmaticConfig.attributeSelectionPattern,
+                attributeSelectionPattern = getAttributeSelectionPattern(specmaticConfig),
```
joelrosario: The `getAttributeSelectionPattern` method which is from the companion object of SpecmaticConfig should not be used here. Instead, given we need the actual attribute selection options, just return the object in a top-level method.

joelrosario (changes requested):

joelrosario (approved):

PR #1589, opened 2025-02-03
Make Sources Field of Specmatic Config Private

sukesh2000:
- Changed the visibility of the sources field in the SpecmaticConfig class from public to private.


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -147,6 +160,56 @@ data class SpecmaticConfig(
...
+                val testPaths = source.getTest() ?: emptyList()
+
+                when (val repository = source.getRepository()) {
+                    null -> GitMonoRepo(testPaths, stubPaths, source.getProvider().toString())
+                    else -> GitRepo(
+                        repository,
+                        source.getBranch(),
+                        testPaths,
+                        stubPaths,
+                        source.getProvider().toString()
```
joelrosario: Instead of `.toString()`, use `.name`.

core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -147,6 +160,56 @@ data class SpecmaticConfig(
         }
     }
 
+    @JsonIgnore
+    fun getGitContracts(): List<ContractSource> {
```
joelrosario: While I see what you're trying to do, config may not be the best place for this, as there's quite a number of classes and construction behaviour unrelated to configuration.

For now can you keep it as it was in loadSources, as this functionality will shortly be impacted by another upcoming PR.

joelrosario (changes requested):

joelrosario (approved):

PR #1590, opened 2025-02-04
Filter excluded endpoints from the html report and coverage report

samyakOO7:
**What**:
Endpoints that are not relevant to specmatic insights are now filtered out in coverage report that is generated.

**Why**:
The config `excludedEndpoints` was used before to filter out endpoints that are not required for specmatic insight reports but as we have filters in place now, we would be moving to filter mechanism to move out endpoints

**How**:
Filter that can identify the endpoints that should be excluded from the OpenAPI coverage report. This can be based on certain criteria, such as path, method type, or tags.

**Checklist**:

1. APIs that are filtered out should not show up in the report (same behaviour as what currently happens with excludedEndpoints) (Done) 
2. Once this is ready, we will start deprecating excludedEndpoints. (Done)
3. --filter​ should be reported in insights. (Pending)


joelrosario (changes requested):
Can you check the coverage report after running the tests on specmatic-order-bff-java before this change (when excludedEndpoints is used in config) and after (by filtering out the /health API) and check if there are any changes in the coverage report.

Ideally the two reports should be identical.

I ask because in this PR the call to `openApiCoverageReportInput.addExcludedAPIs` has been removed. I think the result may be that APIs to be excluded, like '/health', may appear to be missing in the spec. You should be able to confirm this with the testing that I suggested.

junit5-support/src/main/kotlin/io/specmatic/test/SpecmaticJUnitSupport.kt
```diff
@@ -318,7 +318,14 @@ open class SpecmaticJUnitSupport {
                     Pair(tests, endpoints)
                 }
             }
-            openApiCoverageReportInput.addEndpoints(allEndpoints)
+            val testPaths: List<Endpoint> = testScenarios.map {
```
joelrosario: This is not the right approach. `testScenarios` is a sequence, a lazy list, intended to be evaluated one at a time while running tests. You'll need to find another way. One option to consider is to run the filter against the paths in all the available scenarios before the tests are generated, identify the endpoints that remain, and figure out that way which ones were missing, which you could then pass to `openApiCoverageReportInput.addExcludedApis`.

joelrosario (changes requested):
dsfs

junit5-support/src/main/kotlin/io/specmatic/test/SpecmaticJUnitSupport.kt
samyakOO7: Made changes!

joelrosario (changes requested):
Make sure that APis "missing from spec" are also filtered out. They should not appear in the either the coverage report or the HTML report.

core/src/main/kotlin/io/specmatic/core/filters/Filter.kt
```diff
@@ -0,0 +1,84 @@
...
+
+import com.ezylang.evalex.Expression
+import com.ezylang.evalex.config.ExpressionConfiguration
+import io.specmatic.core.Scenario
+import io.specmatic.test.TestResultRecord
+
+data class Filter<T>(
+    val expression: Expression? = null
+) {
+    fun isSatisfiedBy(metadata: T): Boolean {
```
joelrosario: No need to make this a generic parameter

core/src/main/kotlin/io/specmatic/core/filters/Filter.kt
```diff
@@ -0,0 +1,84 @@
...
+        var expression = expression ?: return true
+
+        val variables = when (metadata) {
+            is ScenarioMetadata -> {
```
joelrosario: metadata could just be Any without using Generics. However, I believe there should be a better way of doing this filter. Perhaps introduce a notehr method that accepts the map below, use that by default, and introduce wrapper that uses it for the TestResultRecord. At worst, simply introduce an overloaded method that handles TestResultRecord directly.

joelrosario (changes requested):

joelrosario (approved):

PR #1594, opened 2025-02-05
Add multi port support in stub

yogeshnikam671:
**What**:
Add support to serve requests from multiple ports in stub.
With this feature, a stub specific to a specific spec or list of specs can be started on a specific port.
So potentially we can have multiple ports serving endpoints from different specs.

To configure such a stub, Specmatic configuration v3 is also introduced which roughly looks like follows -
```
version: 3
contracts:
  - filesystem:
      directory: "."
    consumes:
      - imported_product/imported_product.yaml  # port will be defaulted to whatever the stub's default port is (e.g. 9000) 
      - port: 9001
        specs:
          - exported_product/exported_product.yaml
      - port: 9002
        specs:
          - another_exported_product/exported_product.yaml
```

**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

Note - Documentation is pending and needs to be added.


core/src/main/kotlin/io/specmatic/core/config/v3/Consumes.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.JsonNode
+
+sealed class Consumes {
+    data class StringValue(@get:JsonValue val value: String) : Consumes()
+    data class ObjectValue(val specs: List<String>, val port: Int) : Consumes()
+}
+
+class ConsumesDeserializer : JsonDeserializer<List<Consumes>>() {
```
joelrosario, line 15: What's the need for the deserializer?

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
```diff
@@ -360,11 +329,60 @@ class HttpStub(
...
+        keyData: KeyData?,
+        host: String,
+        port: Int
+    ) {
         when (keyData) {
-            null -> connector {
-                this.host = host
-                this.port = port
-            }
+            null -> connectors.addAll(
```
joelrosario, line 374: Has the functionality of running https URLs been tested after this change?

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
```diff
@@ -377,21 +395,46 @@ class HttpStub(
...
+
+    fun serveStubResponse(
+        httpRequest: HttpRequest,
+        port: Int,
+        defaultPort: Int
+    ): HttpStubResponse {
+        return getHttpResponse(
+            httpRequest = httpRequest,
+            features = features,
+            threadSafeStubs = threadSafeHttpStubs.portToListOfStubsMap(defaultPort)[port] ?: threadSafeHttpStubs,
```
joelrosario: What if you pushed the port dereferencing also down one level? It would make these two lines a little easier to read / grok.

core/src/main/kotlin/io/specmatic/stub/ThreadSafeListOfStubs.kt
```diff
@@ -53,7 +67,6 @@ class ThreadSafeListOfStubs(private val httpStubs: MutableList<HttpStubData>) {
             result is Result.Success
         } ?: return null
 
-        this.remove(queueMock)
```
joelrosario, line 56: Why was this line moved outside? If it was about single responsibility of the method, could you rename the method appropriately? The idea is that once a result is found, it should never be found again, and this method will take care of that.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/config/v3/Consumes.kt
yogeshnikam671: Since consumes can be a list of string or a list of objects, the deserializer is required to properly do the required mapping based on the kind of consumes entry (string/list).

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
yogeshnikam671: No.. https has not been tested and I think has not been implemented too. It is a miss I believe. I'll get to it.

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
yogeshnikam671: You mean the portToListOfStubsMap could accept the param port and return the respective stub?

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
yogeshnikam671: Fixed and tested.

core/src/main/kotlin/io/specmatic/stub/ThreadSafeListOfStubs.kt
yogeshnikam671: Since this updates the copy of threadSafeHttpStubsQueue instead of the original one, I had to remove it from here and extract it out.

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
yogeshnikam671: Resolved

joelrosario (approved):

PR #1595, opened 2025-02-05
Fix the issue related to missing branch field while deserializing version 2 of the config

sukesh2000:
When the branch was missing, certain functionality would throw an exception, effectively making the branch mandatory despite it being optional in the config code. This bug has been fixed in this PR.


core/src/test/kotlin/io/specmatic/core/config/SpecmaticConfigAllTest.kt
```diff
@@ -146,4 +151,57 @@ internal class SpecmaticConfigAllTest {
 
         assertThat(parsedJSON(contractsJson)).isEqualTo(parsedJSON(expectedContractsJson))
     }
+
```
joelrosario, line 154: This needs tests checking that the config can be deserialized when `consumes` or `provides` are absent.

joelrosario (changes requested):

joelrosario (approved):

PR #1605, opened 2025-02-10
Make source field of Specmatic Config nullable 

sukesh2000:
- Made the source field nullable
- Moved defaults for source field and it's children to wrapper methods


core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
```diff
@@ -516,7 +514,7 @@ data class SpecmaticConfig(
 
     @JsonIgnore
     fun stubContracts(relativeTo: File = File(".")): List<String> {
```
joelrosario: This is not being used anywhere in the main code, and the logic seems buggy. It is only being used in a test. Remove it from here, and let the logic remain only in the test.

core/src/main/kotlin/io/specmatic/core/config/v2/ContractConfig.kt
```diff
@@ -49,7 +50,11 @@ data class ContractConfig(
     }
 
     fun transform(): Source {
-        return this.contractSource?.transform(provides, consumes) ?: Source(test = provides, stub = consumes)
+        return this.contractSource?.transform(provides, consumes) ?: Source(
+            provider = filesystem,
```
joelrosario: `provider` is now nullable. I believe this line will result in the value `filesystem` appearing during a config upgrade if it was not previously there, when v2 is being upgraded to a future v3. I also believe provider=null is handled in `SpecmaticConfig.loadSources`.

Is there any reason this needs to be initialized to a default value right now? If not, please remove this line.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
sukesh2000: Done, moved the logic to test

core/src/main/kotlin/io/specmatic/core/config/v2/ContractConfig.kt
sukesh2000: Yes, I've removed this line

PR #1616, opened 2025-02-13
Fix flaky test in stateful stub

yogeshnikam671:


core/src/test/kotlin/io/specmatic/stub/stateful/StatefulHttpStubTest.kt
```diff
@@ -661,6 +663,18 @@ class StatefulHttpStubConcurrencyTest {
         latch.await()
         executor.shutdown()
```
joelrosario, line 666: Did you try `executor.awaitTermination()` before calling shutdown? If it works as the documentation suggests, the `waitForProducts` function might not be needed.

joelrosario (changes requested):

core/src/test/kotlin/io/specmatic/stub/stateful/StatefulHttpStubTest.kt
yogeshnikam671: I'll try this

yogeshnikam671: Tested this, it works. I've pushed the change.

PR #1620, opened 2025-02-14
Fix the circular reference issue in BCC and add support to throw parse errors if there are any in the changed specs

yogeshnikam671:
**Checklist**:

- [x] Tests
- [ ] Sonar Quality Gate

### DIFF
--- a/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
+++ b/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
@@ -105,20 +105,24 @@ abstract class BackwardCompatibilityCheckBaseCommand : Callable<Unit> {
     }
 
     private fun checkIfTheChangedSpecsAreValidOrExit(parseResultsOfFilesChangedInCurrentBranch: Set<ParseResult>) {
-        if (parseResultsOfFilesChangedInCurrentBranch.any { it.errorMessages.isNotEmpty() }) {
-            logger.log("The following changed specs are invalid: $newLine")
-            var index = 1
-            parseResultsOfFilesChangedInCurrentBranch.forEach { parseResult ->
-                if (parseResult.errorMessages.isNotEmpty()) {
-                    logger.log("$index. ${parseResult.specPath}, reason(s):")
-                    parseResult.errorMessages.forEachIndexed { idx, errorMessage ->
-                        logger.log("$ONE_INDENT${idx.inc()}. $errorMessage")
-                    }
-                    index++
+        if (parseResultsOfFilesChangedInCurrentBranch.none { it.errorMessages.isNotEmpty() })
+            return
+
+        logger.log("The following changed specs are invalid: $newLine")
+
+        parseResultsOfFilesChangedInCurrentBranch.forEachIndexed { index, parseResult ->
+            if (parseResult.errorMessages.isNotEmpty()) {
+                val ordinal = index + 1
+
+                logger.log("$ordinal. ${parseResult.specPath}, reason(s):")
+
+                parseResult.errorMessages.forEachIndexed { idx, errorMessage ->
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -105,7 +105,15 @@ class OpenApiSpecification(
         }
 
         fun isParsable(openApiFilePath: String): Boolean {
-            return OpenAPIV3Parser().read(openApiFilePath, null, resolveExternalReferences()) != null
+            val options = ParseOptions().apply { isResolve = true; isResolveFully = true }
```
joelrosario, line 108: In the previous version of this function, there were `isResolveRequestBody = true` and `it.isResolveResponses = true`. Here there is `isResolveFully = true`. Is this one a replacement for the other two, and if so has this been tested?

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -105,7 +105,15 @@ class OpenApiSpecification(
...
+            val parseResult = OpenAPIV3Parser().readLocation(openApiFilePath, null, options)
+
+            return parseResult.messages.isEmpty()
```
joelrosario, line 111: What's the significance of checking the message length vs the previous check? I'm guessing messages are indicators from the parser that something isn't quite right. Are all messages an indicator that parsing should break?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -78,9 +78,14 @@ abstract class BackwardCompatibilityCheckBaseCommand : Callable<Unit> {
         val filesChangedInCurrentBranch = getChangedSpecsInCurrentBranch().filter {
             it.contains(targetPath)
         }.toSet()
+        checkIfTheChangedSpecsAreValidOrExit(filesChangedInCurrentBranch.parseResults())
```
joelrosario, line 81: Some users have broken specifications up into small fragments for the purpose of reuse. IIRC the smaller files may have the `.yaml` extension but may not have enough OpenAPI structure to be considered valid OpenAPI specs. If those smaller files change, the backward compatibility check may cut out here too early. Please correct me if I'm understanding this wrong.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
yogeshnikam671: Yes, this is a replacement for the other two and we have done a bit of testing using the specmatic-order-contracts repo.

yogeshnikam671: Yes, this is a replacement for the other two and we have done a bit of testing using the specmatic-order-contracts repo.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
yogeshnikam671: Yes, all messages are an indicator that there is some error in the parsing. So if all messages are empty that means it is a valid parsable openAPI spec.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: A file that just has components and is being used via references in other spec files is still considered a valid OpenAPI structure as per this logic. We have tested this against the common.yaml in specmatic-order-contracts.

PR #1738, opened 2025-04-04
Added support for email format in the contract test

kaushall:
Specmatic currently does not support `email` format specified in the contract, this PR will add support for email format in the contract.


joelrosario (changes requested):
You've added a good test.

We also need however a test ensuring that the (email) type of field gets picked up when it is in an example, and served. You can take inspiration from tests that use the `createStub` function when writing this test.

core/src/test/kotlin/io/specmatic/core/pattern/EmailPatternTest.kt
```diff
@@ -40,6 +46,38 @@ class EmailPatternTest {
...
+                HttpRequest(
+                "GET",
+                "/pets/1"
+                )
+            )
+
+            assertThat(response.status).isEqualTo(200)
+
+            val body = (response.body as JSONObjectValue).jsonObject
+            assertThat(body.keys).containsExactlyInAnyOrder("id", "name", "type", "status", "email")
```
joelrosario, line 109: Should also validate that the returned email is of the right format.

core/src/test/kotlin/io/specmatic/core/pattern/EmailPatternTest.kt
```diff
@@ -1,12 +1,18 @@
...
+import io.specmatic.mock.ScenarioStub
+import io.specmatic.stub.HttpStub
 import io.specmatic.core.value.*
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Tag
 import org.junit.jupiter.api.Test
+import java.io.File
 
 class EmailPatternTest {
     @Test
```
joelrosario, line 18: Can you also add tests for `matches()` method (success / failure) and `newBasedOn()` method.

joelrosario (changes requested):

core/src/test/kotlin/io/specmatic/core/pattern/EmailPatternTest.kt
kaushall: Done

core/src/test/kotlin/io/specmatic/core/pattern/EmailPatternTest.kt
kaushall: This one you already did, I have added a (-)ve test.

joelrosario (approved):

PR #1742, opened 2025-04-08
Added lifecycle hooks for enhanced workflow customization

mik047:
Introduced the **lifecycle hooks** feature to the project, providing enhanced options for customizing and automating specific stages in the workflow. The primary focus of this feature is to allow developers to implement pre-defined hooks at critical stages of the application's lifecycle, facilitating greater flexibility and control.

#### **Key Changes:**
- Added a new `LifecycleHooks` interface to define hooks for key lifecycle events.
- Added hooks for `example validate` command


core/src/main/kotlin/io/specmatic/core/lifecycle/LifecycleHooks.kt
```diff
@@ -0,0 +1,5 @@
+package io.specmatic.core.lifecycle
+
+object LifecycleHooks {
+    var afterLoadingStaticExamples: AfterLoadingStaticExamplesHooks = AfterLoadingStaticExamplesHooks()
```
joelrosario: This should be a `val`, instead of `var`.

joelrosario (changes requested):

joelrosario (approved):

core/src/main/kotlin/io/specmatic/core/lifecycle/LifecycleHooks.kt
mik047: Done!

PR #1756, opened 2025-04-14
Removed spring dependency from `test` and `validate` code

mik047:
Removed spring dependency from 

* `test` command
** main command and test code 
* `examples validate` command
** test code

Removed an unused variable from TestCommand

### DIFF
--- a/application/src/test/kotlin/application/ExamplesCommandTest.kt
+++ b/application/src/test/kotlin/application/ExamplesCommandTest.kt
@@ -1,8 +1,11 @@
 package application
 
+import io.specmatic.core.lifecycle.AfterLoadingStaticExamples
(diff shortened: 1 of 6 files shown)


application/src/test/kotlin/application/ExamplesCommandTest.kt
```diff
@@ -341,16 +341,23 @@ paths:
             No example files found in $examplesDir
             """.trimIndent())
         }
+    }
+
+    @Nested
+    inner class ValidateLifeCycleTests {
+        init {
+            registerTestHook()
```
StarKhan6368: Consider `@BeforeEach` and `@AfterEach` or `@BeforeAll` and `@AfterAll` to register and unregister hook

joelrosario (approved):

PR #1764, opened 2025-04-18
Fixed issue with loading and caching array in virtual service

kaushall:
This PR aims to fix issue with loading and caching array inside response body of `virtual-service` command 

Below is the exception,

```An error occurred while starting the virtual service: class io.specmatic.core.value.JSONObjectValue cannot be cast to class io.specmatic.core.value.JSONArrayValue (io.specmatic.core.value.JSONObjectValue and io.specmatic.core.value.JSONArrayValue are in unnamed module of loader 'app')```


harikrishnan83 (approved):
Looks good, we need to address the structural duplication that is happening with the repeated checks for "JSONArrayValue, JSONObjectValue and else". However we can take that up as a separate effort. cc @joelrosario

PR #1777, opened 2025-04-24
Added spec validation ruleset for virtual service

kaushall:
Added spec validation for virtual service, this PR will ensure the pre-requisites are met before starting the virtual service.

### DIFF
--- a/application/src/main/kotlin/application/VirtualServiceCommand.kt
+++ b/application/src/main/kotlin/application/VirtualServiceCommand.kt
@@ -9,7 +9,6 @@ import io.specmatic.core.Scenario
 import io.specmatic.core.log.StringLog
 import io.specmatic.core.log.consoleLog
 import io.specmatic.core.log.logger
-import io.specmatic.core.pattern.DeferredPattern
 import io.specmatic.core.pattern.JSONObjectPattern
 import io.specmatic.core.utilities.*
 import io.specmatic.mock.ScenarioStub
@@ -93,9 +92,9 @@ class VirtualServiceCommand  : Callable<Int> {
 
         val validateSpec = virtualServiceValidationRuleset(stubData.map{it.first}.flatMap{it.scenarios})
 
-        if (validateSpec.isNotEmpty()) {
+        validateSpec.takeIf{it.isNotEmpty()}?.let{
             logger.log("\n\nThe following errors were found in the specifications:")
-            validateSpec.forEach { logger.log(it) }
+            logger.log(it.joinToString(System.lineSeparator()))
             return
         }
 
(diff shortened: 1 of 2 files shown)


application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -103,21 +114,40 @@ class VirtualServiceCommand  : Callable<Int> {
...
 
+                // Check if the scenario is a POST with 2xx then it should contain an id in the response
+                if (s.httpRequestPattern.method?.uppercase() == "POST" && s.isA2xxScenario()) {
+                    val responsePattern = when (val body = s.httpResponsePattern.body) {
+                        is DeferredPattern -> s.patterns[body.pattern]
+                        is JSONObjectPattern -> body
+                        else -> null
+                    }
+
+                    if (responsePattern is JSONObjectPattern && !responsePattern.pattern.keys.contains("id")) {
```
joelrosario, line 145: The id may be marked optional in the spec, in which case it will appear as "id?". If so, this too needs to be flagged as an error, and a test added for it.

application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -103,21 +114,40 @@ class VirtualServiceCommand  : Callable<Int> {
...
+                        errors.add("Operation: ${s.apiDescription}, does not contains <id> key in the response section as a required field")
+                    }
+                }
 
+                // Check if path contains nested resources
+                val pathSegments = s.path.split("/").filter { it.isNotEmpty() }
+                if (pathSegments.size > 1) {
+                    val potentialId = pathSegments[1]
```
joelrosario, line 157: Can you sync up with both Hari and I on this. This check may need to be slightly more comprehensive.

application/src/test/kotlin/application/VirtualServiceCommandTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    fun `returns errors when POST response does not contain id key`() {
+        val scenario: List<Scenario> = listOf(
+            mockk {
+                every { path } returns "/resource"
+                every { httpRequestPattern.method } returns "POST"
+                every { isA2xxScenario() } returns true
+                every { httpResponsePattern.body } returns mockk<JSONObjectPattern> {
```
joelrosario: Can you put an actual JSONObjectPattern here please

application/src/test/kotlin/application/VirtualServiceCommandTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        )
+        val errors = virtualServiceValidationRuleset(scenario)
+        assertEquals(1, errors.size)
+        assertEquals("Invalid HTTP method TRACE in path /resource, The supported methods are GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS", errors[0])
+    }
+
+    @Test
+    fun `returns errors when POST response does not contain id key`() {
+        val scenario: List<Scenario> = listOf(
+            mockk {
```
joelrosario: Was it too complicated to instantiate a Scenario?

application/src/test/kotlin/application/VirtualServiceCommandTest.kt
```diff
@@ -0,0 +1,75 @@
...
+                every { isA2xxScenario() } returns true
+                every { httpResponsePattern.body } returns mockk<JSONObjectPattern> {
+                    every { pattern.keys } returns setOf("id")
+                }
+                every { apiDescription } returns "POST /resource/add"
+            }
+        )
+        val errors = virtualServiceValidationRuleset(scenario)
+        assertEquals(1, errors.size)
+        assertEquals("Operation POST /resource/add, contains invalid nested resource add, Resources should follow flat structure like /resource or /resource/{id}", errors[0])
```
joelrosario: This may need to change depending depending on whether the validation for URL ids is modified per PR comments.

joelrosario (changes requested):

application/src/main/kotlin/application/VirtualServiceCommand.kt
kaushall: Yes, I have considered that case, if you read the error message it says `does not contains <id> key in the response section as a required field`.

application/src/test/kotlin/application/VirtualServiceCommandTest.kt
kaushall: Nope, I wanted to mock some specific behaviour for the unit testing.

application/src/main/kotlin/application/VirtualServiceCommand.kt
kaushall: Rule 1:
POST - 1 Segment
GET - 1/2 Segments
OTHERS - 2 Segments

Rule 2:
When 2 segments check for the second segment, it should be number/string.

application/src/main/kotlin/application/VirtualServiceCommand.kt
joelrosario: Actually, I wanted to be explicit and say, "id" is given as an optional key, it needs to be mandatory. However, let's address the code structure first.

application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -103,21 +114,35 @@ class VirtualServiceCommand  : Callable<Int> {
...
-
-
-
-
-
-
 
+    companion object {
+        fun virtualServiceValidationRuleset(scenarios: List<Scenario>): MutableList<String> {
+            val errors: MutableList<String> = mutableListOf()
```
joelrosario: Use an immutable List instead of MutableList

application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -103,21 +114,35 @@ class VirtualServiceCommand  : Callable<Int> {
...
+            val supportedMethods = setOf("GET", "POST", "PUT", "DELETE", "PATCH", "HEAD", "OPTIONS")
+            scenarios.forEach{ s ->
```
joelrosario: Use map instead of forEach

application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -103,21 +114,35 @@ class VirtualServiceCommand  : Callable<Int> {
...
+                // Check if the scenario is a valid HTTP method
+                if(s.httpRequestPattern.method !in supportedMethods){
```
joelrosario: This line of code (`s.httpRequestPattern.method`) breaks the "law of demeter". Read up on this and fix the issue. Same for other places a similar dot-dot-dot pattern is found.

application/src/main/kotlin/application/VirtualServiceCommand.kt
```diff
@@ -88,6 +91,14 @@ class VirtualServiceCommand  : Callable<Int> {
             false
         )
 
+        val validateSpec = virtualServiceValidationRuleset(stubData.map{it.first}.flatMap{it.scenarios})
```
joelrosario: `validateSpec` should be a List<String>, rather than a MutableList.

joelrosario (changes requested):

application/src/main/kotlin/application/VirtualServiceCommand.kt
kaushall: As discussed, no need to take this up for now.

application/src/test/kotlin/application/VirtualServiceCommandTest.kt
kaushall: As discussed, no need to take this up for now.

joelrosario (approved):

PR #1780, opened 2025-04-25
Replaced regex based expression builder with token based expression builder in the filters

kaushall:
- Replaced regex based expression builder with token based expression builder in the filters
- Made filter names and usage consistent across the filters


core/src/main/kotlin/io/specmatic/core/filters/ScenarioMetadata.kt
```diff
@@ -17,6 +17,6 @@ data class ScenarioMetadata(
             .with(ScenarioFilterTags.STATUS_CODE.key, statusCode.toString())
             .with(ScenarioFilterTags.HEADER.key, header.joinToString(","))
             .with(ScenarioFilterTags.QUERY.name, query.joinToString(","))
```
kaushall, line 19: Make it consistent and use `key`

ketan (approved):
LGTM. @joelrosario  - could you comment on removal of the test for silently ignoring wrong syntax?

PR #1789, opened 2025-05-01
Updated email regex to filter emails with invalid domains

kaushall:
- Updated email regex to filter emails with invalid domains
- Limit the quantifiers to prevent uncontrolled expansion of the `*` operator
- Added character limits for local and domain parts as per RFC 5322 compliance

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/pattern/EmailPattern.kt
+++ b/core/src/main/kotlin/io/specmatic/core/pattern/EmailPattern.kt
@@ -6,7 +6,7 @@ import io.specmatic.core.value.StringValue
 import io.specmatic.core.value.Value
 import java.util.*
 
-private const val EMAIL_REGEX = "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}\$"
+private const val EMAIL_REGEX = "^[a-zA-Z0-9](?:[\\w.-]{0,62}[a-zA-Z0-9])?@[a-zA-Z0-9](?:[a-zA-Z0-9.-]{0,253}[a-zA-Z0-9])?\\.[a-zA-Z]{2,}\$"
 
 class EmailPattern (private val stringPatternDelegate: StringPattern) :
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/io/specmatic/core/pattern/EmailPattern.kt
```diff
@@ -6,7 +6,7 @@ import io.specmatic.core.value.StringValue
 import io.specmatic.core.value.Value
 import java.util.*
 
-private const val EMAIL_REGEX = "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}\$"
+private const val EMAIL_REGEX = "^[a-zA-Z0-9](?:[\\w.-]{0,62}[a-zA-Z0-9])?@[a-zA-Z0-9](?:[a-zA-Z0-9.-]{0,253}[a-zA-Z0-9])?\\.[a-zA-Z]{2,}\$"
```
joelrosario, line 9: What would be a good way to eliminate this regular expression?

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/pattern/EmailPattern.kt
ketan: Suggested read - https://stackoverflow.com/q/201323

ketan: Also this https://docs.oracle.com/javaee/7/api/javax/mail/internet/InternetAddress.html

kaushall: @joelrosario This regex is safe and accurate for common email validations/generations but doesn't support quoted local parts or domain literals (like [192.168.1.1]) as per RFC5322. But if you want a complete RFC 5322 compliant validator/generator then we should use some library as Ketan suggested above.

PR #1805, opened 2025-05-08
fix(vs): load and cache inline examples data in the virtual service

kaushall:
This PR will load and cache inline examples data in the virtual service

### DIFF
--- a/core/src/main/kotlin/io/specmatic/stub/stateful/StatefulHttpStub.kt
+++ b/core/src/main/kotlin/io/specmatic/stub/stateful/StatefulHttpStub.kt
@@ -5,9 +5,7 @@ import io.ktor.server.application.*
 import io.ktor.server.engine.*
 import io.ktor.server.http.content.*
 import io.ktor.server.netty.*
-import io.ktor.server.plugins.cors.*
 import io.ktor.server.plugins.cors.CORS
-import io.ktor.server.plugins.cors.routing.*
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/io/specmatic/stub/stateful/StatefulHttpStub.kt
```diff
@@ -723,67 +721,74 @@ class StatefulHttpStub(
...
-                    idValue = idValueFor(DEFAULT_CACHE_RESPONSE_ID_KEY, responseBody)
+                    idValue = idValueFor(DEFAULT_CACHE_RESPONSE_ID_KEY, body)
                 )
             }
+            return
         }
 
-        return stubCache
+        if (responseBody !is JSONObjectValue) return
+        if (request.method == "POST" && request.body !is JSONObjectValue) return
```
ketan, line 774: Use `return` in a block:

```
if(...) {
  return ...
}
```

ketan (commented):
For @joelrosario to take a look. Minor nitpicks in my review.

core/src/main/kotlin/io/specmatic/stub/stateful/StatefulHttpStub.kt
kaushall: Thanks for the feedback @ketan, Just curious, what is the rationale behind putting it inside the block?

PR #1820, opened 2025-05-14
Fix path matching logic for shadowed paths

StarKhan6368:
**What**: Fix path matching logic for shadowed paths

**Why**: 
- If the `OAS` has overlapping paths, such as `\test\{id}` and `\test\latest`, without any other unique identifier, the examples intended solely for `\test\latest` may be loaded into `\test\{id}` later, resulting in load-time failures due to mismatching body or other values
- Similarly when running a stub server, requests for `\test\latest` may unexpectedly return response associated with `\test\{id}`

**How**:
- Conflicting values are pre-calculated for each path segment and are later used in pattern creation
- When matching a path to the specified pattern, the value is checked against the conflicts

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)


core/src/main/kotlin/io/specmatic/core/HttpPathPattern.kt
```diff
@@ -79,12 +79,12 @@ data class HttpPathPattern(
...
 
         val failures = results.filterIsInstance<Failure>()
-
         val finalMatchResult = Result.fromResults(failures)
-
-        val failureReason = if (structureMatches(path, resolver)) {
-            FailureReason.URLPathParamMismatchButSameStructure
-        } else FailureReason.URLPathMisMatch
+        val failureReason = when {
+            finalMatchResult is Failure && finalMatchResult.hasReason(FailureReason.SegmentConflict) -> FailureReason.SegmentConflict
```
joelrosario: Do you think it might make sense to check for structure match before segment match (because if structure match si wrong then segment match may not be relevant)?

core/src/main/kotlin/io/specmatic/core/HttpPathPattern.kt
```diff
@@ -58,7 +58,7 @@ data class HttpPathPattern(
             try {
 
                 val parsedValue = urlPathPattern.tryParse(token, resolver)
-                val result = resolver.matchesPattern(urlPathPattern.key, urlPathPattern.pattern, parsedValue)
+                val result = urlPathPattern.matches(parsedValue, resolver)
```
joelrosario, line 61: resolver.matchesPattern also checks in case the value is a pattern token. Do you think changing from resolver.matchesPattern to urlPathPattern.matches will have repercussions?

StarKhan6368: The method URLPathSegmentPattern.matches internally calls resolver.matchesPattern, and it will now also perform a conflicts check

core/src/main/kotlin/io/specmatic/core/HttpPathPattern.kt
StarKhan6368: That makes sense, thanks, updated the code to do the same

PR #1826, opened 2025-05-16
Implement support for logical AND in security schemes

### LINKED ISSUE #932: Secure schema not respected when generating test scenarios
**Description**
According to the [spec](https://spec.openapis.org/oas/v3.0.0#security-requirement-object)  and the Swagger [documentation](https://swagger.io/docs/specification/authentication/)
>That is, security is an array of hashmaps, where each hashmap contains one or more named security schemes. Items in a hashmap are combined using logical AND, and array items are combined using logical OR. 

For instances
```
security:
  - apiKey1: []
    apiKey2: []
```

Should generate only one test with two required headers, while specmatic generates two separate tests for each headers which is wrong. Specmatic behaviour is correct only for the following security configuration

```
security:
  - apiKey1: []
  - apiKey2: []
```

**Steps to reproduce**
I wrote an example [test](https://github.com/znsio/specmatic/pull/931/files)

**Expected behavior**
Create one test scenario when logical AND used

**Screenshots**
N/A

**System Information:**
N/A

StarKhan6368:
**What**: Implement support for logical AND in security schemes

**Why**:
- Previously, we only supported `OR` between the security schemes defined in the `OAS`.
- However, a path can define multiple security schemes according to the OAS standards.

**Sources**:
- https://swagger.io/docs/specification/v3_0/authentication/#using-multiple-authentication-types
- https://spec.openapis.org/oas/v3.1.0.html#security-requirement-object

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)
- [ ] Roadmap updated (share link)

**Issue ID**:
Closes: #932

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
@@ -32,4 +32,8 @@ data class APIKeyInHeaderSecurityScheme(val name: String, private val apiKey:Str
     }
 
     override fun isInRow(row: Row): Boolean = row.containsField(name)
+
+    override fun isInRequest(request: HttpRequest): Boolean {
+        return request.headers.containsKey(name)
+    }
 }
--- a/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
@@ -34,6 +34,6 @@ data class APIKeyInHeaderSecurityScheme(val name: String, private val apiKey:Str
     override fun isInRow(row: Row): Boolean = row.containsField(name)
 
     override fun isInRequest(request: HttpRequest, complete: Boolean): Boolean {
-        return request.headers.containsKey(name)
+        return request.hasHeader(name)
     }
 }
--- a/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
@@ -33,7 +33,7 @@ data class APIKeyInHeaderSecurityScheme(val name: String, private val apiKey:Str
 
     override fun isInRow(row: Row): Boolean = row.containsField(name)
 
-    override fun isInRequest(request: HttpRequest): Boolean {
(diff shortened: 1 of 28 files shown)


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -974,21 +986,16 @@ class OpenApiSpecification(
...
-                    securityScheme
-                }.ifEmpty {
-                    mapOf(NO_SECURITY_SCHEMA_IN_SPECIFICATION to NoSecurityScheme())
-                }
+        val securitySchemeEntries = parsedOpenApi.components?.securitySchemes.orEmpty()
+        val securitySchemes = securitySchemeEntries.entries.associate { (schemeName, scheme) ->
+            schemeName to toSecurityScheme(schemeName, scheme)
+        }
+
+        val securitySchemesForRequestPattern = parseOperationSecuritySchemas(operation) { name ->
```
joelrosario: Why not pass the securitySchemes directly instead of a function?

core/src/main/kotlin/io/specmatic/conversions/APIKeyInHeaderSecurityScheme.kt
```diff
@@ -32,4 +32,8 @@ data class APIKeyInHeaderSecurityScheme(val name: String, private val apiKey:Str
     }
 
     override fun isInRow(row: Row): Boolean = row.containsField(name)
+
+    override fun isInRequest(request: HttpRequest, complete: Boolean): Boolean {
```
joelrosario, line 36: maybe this could be request.containsHeader(name) so that we don't use the headers property directly.

core/src/test/kotlin/io/specmatic/core/RunContractTestsUsingScenario.kt
```diff
@@ -768,9 +765,15 @@ paths:
...
-                    val expectedValue = extractValue(request)
-                    assertThat(actualValue).isEqualTo(expectedValue)
+                    val schemesToCheck = when(securitySchema) {
+                        is CompositeSecurityScheme -> securitySchema.schemes
+                        else -> listOf(securitySchema)
+                    }
+                    assertThat(schemesToCheck).allSatisfy {
+                        val actualValue = extractValue(request)
+                        val expectedValue = extractValue(request)
+                        assertThat(actualValue).isEqualTo(expectedValue)
```
joelrosario, line 775: both the above seem to extract the same value, so won't the assertion just pass?

joelrosario (approved):

PR #1846, opened 2025-05-27
feat(filters): implement consistent filters in test and stub commands

ketan:


ketan:
Depends on https://github.com/znsio/docs.specmatic.io/pull/31

PR #1861, opened 2025-05-29
Calculate and test specificity for values and HTTP requests

mik047:
**What**:  
Added `specificity` calculation logic for `Value` subclasses and HTTP requests. Refactored test utilities to use JUnit 5's `Stream<Arguments>` for better maintainability.

**Why**:  
To enhance comparison precision for paths, headers, query parameters, and bodies in HTTP requests, and to improve test code clarity and compliance.

**How**:  
- Introduced `specificity` methods to `Value` subclasses.  
- Implemented specificity score logic in `HttpRequest`.  
- Migrated parameterized tests to use `Stream<Arguments>`.  
- Added comprehensive unit tests for all changes.

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/value/JSONArrayValue.kt
+++ b/core/src/main/kotlin/io/specmatic/core/value/JSONArrayValue.kt
@@ -98,4 +98,8 @@ data class JSONArrayValue(override val list: List<Value>) : Value, ListValue, JS
     override fun generality(): Int {
         return list.sumOf { it.generality() }
     }
+
+    override fun specificity(): Int {
+        return list.sumOf { it.specificity() }
+    }
 }
--- a/core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
+++ b/core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
@@ -124,6 +124,10 @@ data class JSONObjectValue(val jsonObject: Map<String, Value> = emptyMap()) : Va
     override fun generality(): Int {
         return jsonObject.values.sumOf { it.generality() }
     }
+
+    override fun specificity(): Int {
+        return jsonObject.values.sumOf { it.specificity() }
+    }
 }
 
 internal fun dictionaryToDeclarations(jsonObject: Map<String, Value>, types: Map<String, Pattern>, exampleDeclarations: ExampleDeclarations): Triple<Map<String, DeferredPattern>, Map<String, Pattern>, ExampleDeclarations> {
--- a/core/src/main/kotlin/io/specmatic/core/value/Value.kt
+++ b/core/src/main/kotlin/io/specmatic/core/value/Value.kt
@@ -41,9 +41,7 @@ interface Value {
         return 0
     }
 
-    fun specificity(): Int {
-        return 0
-    }
+    fun specificity(): Int
 }
 
 fun Value.mergeWith(other: Value): Value {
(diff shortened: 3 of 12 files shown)


core/src/test/kotlin/io/specmatic/core/HttpRequestTest.kt
```diff
@@ -412,8 +383,188 @@ internal class HttpRequestTest {
...
+            Arguments.of("/persons/(string)/1/(string)", 2),
+            Arguments.of("/persons/group/(string)/1/(string)", 2),
+        )
+
+        @JvmStatic
+        fun urlPathToExpectedPathSpecificity(): Stream<Arguments> = Stream.of(
+            Arguments.of(null, 0),
+            Arguments.of("", 1),
+            Arguments.of("/", 1),
+            Arguments.of("/persons", 2),
```
joelrosario, line 471: This should be 1

core/src/main/kotlin/io/specmatic/core/value/JSONArrayValue.kt
```diff
@@ -98,4 +98,8 @@ data class JSONArrayValue(override val list: List<Value>) : Value, ListValue, JS
     override fun generality(): Int {
         return list.sumOf { it.generality() }
     }
+
+    override fun specificity(): Int {
+        return list.sumOf { it.specificity() }
```
joelrosario, line 103: Return 1 if the list is empty

core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
```diff
@@ -124,6 +124,10 @@ data class JSONObjectValue(val jsonObject: Map<String, Value> = emptyMap()) : Va
     override fun generality(): Int {
         return jsonObject.values.sumOf { it.generality() }
     }
+
+    override fun specificity(): Int {
+        return jsonObject.values.sumOf { it.specificity() }
```
joelrosario, line 129: return 1 if the object is empty

core/src/main/kotlin/io/specmatic/core/value/Value.kt
```diff
@@ -40,6 +40,10 @@ interface Value {
     fun generality(): Int {
         return 0
     }
+
+    fun specificity(): Int {
+        return 0
```
joelrosario: BinaryValue is not handle, and hence will return 0.

Perhaps, remove the defaulter and override everywhere else to return what's relevant. It's a small method and if we're creating a new Value later we need to make a conscious decision about the value to return.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/value/JSONArrayValue.kt
mik047: As discussed and concluded, we would treat empty list as `0`

core/src/main/kotlin/io/specmatic/core/value/JSONObjectValue.kt
mik047: As discussed and concluded, we would treat empty map as `0`

core/src/main/kotlin/io/specmatic/core/value/Value.kt
mik047: made `specifity()` abstract for children to explicitly define relevant values.

Added values for BinaryValue and others where it was deriving default value.

joelrosario (approved):

PR #1910, opened 2025-06-12
retain extensions during allOf resolution

yogeshnikam671:
- [x] Tests

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
@@ -1505,13 +1505,20 @@ class OpenApiSpecification(
                             )
                         }.flatMap { (componentName, properties) ->
                             schemaProperties.map {
-                                componentName to combine(it.properties, properties)
+                                componentName to SchemaProperty(
+                                    extensions = it.extensions,
+                                    properties = combine(it.properties, properties)
+                                )
                             }
                         }
 
                         result
-                    }.flatten().map { (componentName, properties) ->
-                        toJSONObjectPattern(properties, "(${componentName})")
+                    }.flatten().map { (componentName, schemaProperty) ->
+                        toJSONObjectPattern(
+                            schemaProperty.properties,
+                            "(${componentName})",
+                            schemaProperty.extensions
+                        )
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -1492,7 +1505,7 @@ class OpenApiSpecification(
                             )
                         }.flatMap { (componentName, properties) ->
                             schemaProperties.map {
-                                componentName to combine(it, properties)
+                                componentName to combine(it.properties, properties)
```
joelrosario: This block of code handles oneOfs within an allOf. Instead of a single allOf, it will result in multiple objects, one per oneOf within the allOf. We may be missing the logic to handle extensions in this case.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -1511,7 +1524,9 @@ class OpenApiSpecification(
                         )
                     else if(allDiscriminators.isNotEmpty())
                         AnyPattern(
-                            pattern = schemaProperties.zip(allDiscriminators.schemaNames).map { (properties, schemaName) ->
+                            pattern = schemaProperties.map {
+                                it.properties
+                            }.zip(allDiscriminators.schemaNames).map { (properties, schemaName) ->
                                 toJSONObjectPattern(properties, "(${schemaName})")
```
joelrosario: We may be missing adding extensions here

joelrosario (changes requested):
Does the oneOf branch (line 1560) need the logic for handling extensions?

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
yogeshnikam671: Resolved

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
yogeshnikam671: Resolved

PR #1923, opened 2025-06-23
feat(logging): add a function to just log request and response in HttpLogMessage

yogeshnikam671:
- [x] Tests


yogeshnikam671:
The method is not required as the expected functionality can be achieved with existing code.
Thanks @joelrosario. Closing the PR.

PR #1934, opened 2025-06-28
Resolve edge case related to shadowed path matching

StarKhan6368:
**What**: Resolve edge case related to shadowed path matching

**Reason**: 
- When there is a shadowed path with parameters that do not share the same schema, such as `/orders/(number)` and `/orders/bulk`, or `/(string)/orders/(string)` and `/(string)/orders/bulk` the path matching logic fails to provide the correct failure reason. This issue causes the `interactive server` to erroneously duplicate the example for both paths in the UI

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)


StarKhan6368:
Issue resolved by https://github.com/specmatic/specmatic/pull/2016

PR #1944, opened 2025-07-03
chore: allow system properties to affect `HttpClient`

kobim:
**What**:

Allowing [Java's Networking Properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) to be specified and control the behavior of `HttpClient` during the requests.

**Why**:

Some servers might not be directly accessible for testing, and require passing through proxies. Without an option to define a proxy, we end up with spinning reverse proxies for the purpose of testing.

**How**:

The `Apache` engine factory of `ktor` allows us to accept "system properties" (defined via `-Dproperty=value` while running the JAR) that would define the behavior of the `HttpClient`.

As the engine provides sane defaults, none of the existing behavior will change if we toggle this option on.

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Documentation added/updated (share link) **N/A**
- [ ] Sample Project added/updated (share link) **N/A**
- [ ] Demo video (share link) **N/A**
- [ ] Article on Website (share link) **N/A**
- [ ] Roadmpap updated (share link) **N/A**
- [ ] Conference Talk (share link) **N/A**

### DIFF
--- a/core/src/main/kotlin/io/specmatic/test/ApacheHttpClientFactory.kt
+++ b/core/src/main/kotlin/io/specmatic/test/ApacheHttpClientFactory.kt
@@ -25,6 +25,7 @@ class ApacheHttpClientFactory(override val timeoutPolicy: TimeoutPolicy): HttpCl
                         .build()
                 )
                 setSSLHostnameVerifier(NoopHostnameVerifier())
+                useSystemProperties()
             }
         }
 


ketan:
LGTM. Will need to investigate why the builds keep failing. @kaushall - could you investigate please?

joelrosario:
Thanks for the PR @kobim.

harikrishnan83:
Thanks again @kobim , your fix is now available as part of [release 2.16.0](https://github.com/specmatic/specmatic/releases/tag/2.16.0).

PR #1993, opened 2025-08-21
feat: add ability to set SPECMATIC_EXCLUDED_ENDPOINTS through system property as well.

yogeshnikam671:
- [x] Tests

### DIFF
--- a/junit5-support/src/main/kotlin/io/specmatic/test/reports/ReportProcessor.kt
+++ b/junit5-support/src/main/kotlin/io/specmatic/test/reports/ReportProcessor.kt
@@ -2,7 +2,6 @@ package io.specmatic.test.reports
 
 import io.specmatic.core.ReportConfiguration
 import io.specmatic.core.SpecmaticConfig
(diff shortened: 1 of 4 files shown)


junit5-support/src/main/kotlin/io/specmatic/test/reports/ReportProcessor.kt
```diff
@@ -2,10 +2,11 @@ package io.specmatic.test.reports
 
 import io.specmatic.core.ReportConfiguration
 import io.specmatic.core.SpecmaticConfig
+import io.specmatic.test.ContractTestSettings
 import io.specmatic.test.reports.renderers.ReportRenderer
 
 interface ReportProcessor<T> {
-    fun process(specmaticConfig: SpecmaticConfig)
+    fun process(specmaticConfig: SpecmaticConfig, settings: ContractTestSettings)
```
joelrosario: Can you accept settings in the constructor of this object instead of as a parameter.

joelrosario (changes requested):

PR #2014, opened 2025-09-17
AfterAll Hook and New Report Formatter

yogeshnikam671:


joelrosario (approved):

junit5-support/src/main/kotlin/io/specmatic/test/SpecmaticJUnitSupport.kt
```diff
@@ -443,8 +450,11 @@ open class SpecmaticJUnitSupport {
                     throw e
                 } finally {
                     if (testResult != null) {
+                        val durationMillis = Duration.between(startTime, Instant.now()).toMillis()
```
joelrosario: What do you think of putting this duration measurement inside the ScenarioAsTest class itself, so that the duration field in testRecord doesn't have to be updated post-facto with the duration? The value could default to 0, which is what it would be when no tests run. What do you think?

joelrosario (changes requested):

junit5-support/src/main/kotlin/io/specmatic/test/SpecmaticJUnitSupport.kt
yogeshnikam671: Yeah, having the duration measurement within ScenarioAsTest makes sense. But the other point of post-facto updation is unclear. Currently, ScenarioAsTest returns a Pair<Result, HttpResponse> and TestResultRecord is created using them. We'll have to return the duration as part of Result or something. Need some clarification

joelrosario (approved):

PR #2026, opened 2025-10-01
feat: add support for providing constants in dictionary which can be replaced wherever the placeholders are present

yogeshnikam671:


core/src/test/resources/dictionary/dictionary_with_constants.json
```diff
@@ -0,0 +1,36 @@
...
+      "name": "Volleyball",
+      "category": "sports"
+    },
+    "ORDERS": [
+      { "orderId": 1234 },
+      { "orderId": 4567 }
+    ]
+  },
+  "getOrder": {
+    "orderId": "<ORDER_ID>"
```
joelrosario: We use the syntax $(ORDER_ID) for substituting values in examples.

joelrosario (changes requested):

PR #2076, opened 2025-11-13
Fix proxy combining schemas due non-unqiue names generation

StarKhan6368:
**What**:
- Resolve the issue with proxy schema combination due to non-unique name generation.  
- Ensure the proxy correctly manages various request and response content types, as well as status codes, during conversion.

**Checklist**:
- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/Feature.kt
+++ b/core/src/main/kotlin/io/specmatic/core/Feature.kt
@@ -1136,34 +1136,14 @@ data class Feature(
     }
 
     private fun toOpenAPIURLPrefixMap(urls: List<String>): Map<String, String> {
-        val normalisedURL = urls.map { url ->
-            val path =
-                url.removeSuffix("/").removePrefix("http://").removePrefix("https://").split("/").joinToString("/") {
-                    if (it.toIntOrNull() != null)
-                        "1"
-                    else
-                        it
+        return urls.associateWith { url ->
+            url.removeSuffix("/").removePrefix("http://").removePrefix("https://")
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -1136,34 +1136,14 @@ data class Feature(
     }
 
     private fun toOpenAPIURLPrefixMap(urls: List<String>): Map<String, String> {
-        val normalisedURL = urls.map { url ->
```
StarKhan6368, line 1139: @joelrosario, I didn't fully understand the previous logic, open to reverting if it was important or logical

core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -1196,70 +1176,72 @@ data class Feature(
...
+                    pattern = responseBody,
+                    status = scenario.status,
+                    method = scenario.method,
+                    contentType = scenario.httpResponsePattern.headersPattern.contentType,
+                ).let(::withPatternDelimiters)
 
+                val newResponseBody = updateJsonPatternName(responseBody, newTypeName)
+                val newTypes = scenario.patterns[oldTypeName]?.let {
+                    scenario.patterns.minus(oldTypeName).plus(newTypeName to it)
+                } ?: scenario.patterns
```
joelrosario, line 1216: Why is this fallback required, is there a case we may be missing?

PR #2089, opened 2025-11-20
Use DTOs from specmatic-reporter

yogeshnikam671:


core/src/main/kotlin/io/specmatic/stub/report/StubUsageReport.kt
```diff
@@ -17,34 +21,36 @@ class StubUsageReport(
...
-                                    && it.sourceRepositoryBranch == key.sourceRepositoryBranch
-                                    && it.specification == key.specification
-                                    && it.serviceType == key.serviceType
-                        }
-                    )
-                }
+            val operations = recordsOfGroup.groupBy {
+                Triple(it.path, it.method, it.responseCode)
+            }.map { (operationGroup, _) ->
+                HTTPStubUsageOperation(
```
yogeshnikam671: Use setters to set the constructor values to avoid possible breakage if arguments ordering changes in the constructor.

yogeshnikam671 (commented):
Address the comments.

PR #2095, opened 2025-11-24
Improvements to internal patterns

StarKhan6368:
**What**: Improvements to internal patterns

**Why**:
- The primary goal is to provide partial support, if not full support, for OpenAPI 3.1 specification schemas
- Additionally, address consistency and other issues in the implementations

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)


core/src/main/kotlin/io/specmatic/core/pattern/ExactValuePattern.kt
```diff
@@ -40,11 +39,18 @@ data class ExactValuePattern(override val pattern: Value, override val typeAlias
     }
 
     override fun generate(resolver: Resolver) = pattern
-    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(this))
+    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> {
+        return sequenceOf(HasValue(this, "is set to exact value of ${pattern.displayableValue()}"))
```
joelrosario, line 43: Need make sure there's a test on the EnumPattern class ensuring that the valueDetails of newBasedOn values is overridden and this message is not seen. It should only be seen for const.

core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
```diff
@@ -91,7 +91,7 @@ data class AnyPattern(
     override fun jsonObjectPattern(resolver: Resolver): JSONObjectPattern? {
         if (this.hasNoAmbiguousPatterns().not()) return null
 
-        val pattern = this.pattern.first { it !is NullPattern }
+        val pattern = this.pattern.first(::isEmpty)
```
joelrosario: This should be `firstNot`

core/src/main/kotlin/io/specmatic/core/pattern/ExactValuePattern.kt
StarKhan6368: I've added tests to lock-in this behavior, please take a look at
- `io.specmatic.core.pattern.EnumPatternTest.TestGeneration#should retain the correct enum newBasedOn return value details`
- `io.specmatic.core.pattern.EnumPatternTest.TestGeneration#should retain the correct enum negativeBasedOn return value details`

core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
StarKhan6368: Thanks for the catch, fixed this and any other mistakes I've found in AnyPattern

StarKhan6368:
I am closing this pull request since the [dependent PR](https://github.com/specmatic/specmatic/pull/2096) has been merged with the relevant commits included, rather than through a two-PR merge, i.e. this and then following with the dependent

PR #2119, opened 2025-12-08
chore: weave in compatibility information received from the hooks

ketan:


ketan:
Closed. This PR is not relevant anymore

PR #2122, opened 2025-12-09
feat: implement CTRF report generation functionality

ketan:


core/src/main/kotlin/io/specmatic/core/report/CTRFReportGenerator.kt
```diff
@@ -0,0 +1,42 @@
...
+import io.specmatic.core.getConfigFilePath
+import io.specmatic.reporter.ctrf.CtrfReportGenerator
+import io.specmatic.reporter.ctrf.model.CtrfSpecConfig
+import io.specmatic.reporter.reporting.ReportProvider
+import io.specmatic.test.TestResultRecord
+import io.specmatic.test.TestResultRecord.Companion.getCoverageStatus
+import java.io.File
+import java.util.*
+
+object CTRFReportGenerator {
```
yogeshnikam671: In this method technically HTML report is also getting generated.
Should we just name this object ReportGenerator instead of CTRFReportGenerator?
We could later move the console TEXT report (the only remaining report type) generation method as well into this object making this a central place for all kinds of report generation.

yogeshnikam671 (changes requested):

PR #2132, opened 2025-12-16
Coverage Report Fixes

jaydeepk:
This PR fixes the following issues in the generation of the coverage report:
1. The report was not indicating "missing in spec" operations correctly. This was  due to duplicate test records being added for "missing in spec" tests. This is now fixed
2. When calculating coverage percentage, the current logic includes "missing in spec" operations (which causes discrepancies in Insights). Hence, "missing in spec" operations will not be considered while calculating coverage percentage.


StarKhan6368 (approved):
Looks good to me, We'll need to tweak how the studio calculates coverage and possibly remarks @joelrosario , I'll make a note of it so we can sync it up as well

joelrosario (approved):

PR #2136, opened 2025-12-17
fix: fix ctrfSpecConfig generation

yogeshnikam671:


core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
```diff
@@ -40,4 +42,11 @@ object ReportGenerator {
             hook.generateReport(report, reportDir)
         }
     }
+
+    private fun validateCtrfSpecConfigs(specConfigs: List<CtrfSpecConfig>) {
+        if (specConfigs.isEmpty()) throw IllegalArgumentException("CtrfSpecConfigs cannot be empty")
```
joelrosario: I presume if this happens, it means there's a bug. However by throwing an exception, I'm guessing we are terminating the report generation and possibly the test suite itself, which might have passed but will now look like it failed...? Is there a more graceful way of handling this case through a warning log?

joelrosario (approved):

PR #2139, opened 2025-12-22
Fix requestContentType handling in reports

jaydeepk:
This PR includes the following fixes:
1. Inclusion of requestContentType for not covered test result records (which was previously being skipped)
2. requestContentType to be sent  as null wherever it is not applicable or available (instead of passing an empty string)


yogeshnikam671 (approved):
LGTM

joelrosario (approved):

PR #2142, opened 2025-12-23
Implement initial support for Rule-Id violations

StarKhan6368:
**What**: Implement initial support for Rule-Id violations

**Why**: Each violation is now assigned a Rule-Id (e.g., R0001) to enable users to quickly reference its precise definition and suggested resolution in the rules documentation, and allowing the UI to deliver better reporting

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)


core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
```diff
@@ -214,7 +214,8 @@ data class AnyPattern(
 
         val failuresWithUpdatedBreadcrumbs = addTypeInfoBreadCrumbs(matchResults)
 
-        return Result.fromFailures(failuresWithUpdatedBreadcrumbs)
+        val didNotMatchAnyOptionFailure = Failure(message = "Value didn't match any of the schema options", ruleViolation = StandardRuleViolation.ONE_OF_VALUE_MISMATCH)
```
joelrosario: Why is this extra failur added?

core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
```diff
@@ -339,7 +340,7 @@ data class AnyPattern(
         val compatibleResult = otherPattern.fitsWithin(patternSet(thisResolver), otherResolver, thisResolver, typeStack)
 
         return if(compatibleResult is Failure && allValuesAreScalar())
-            mismatchResult(this, otherPattern, thisResolver.mismatchMessages)
+            patternMismatchResult(this, otherPattern, thisResolver.mismatchMessages)
```
joelrosario, line 339: Seems like backward compatibility errors don't have rule ids yet?

core/src/main/kotlin/io/specmatic/core/pattern/DateTimePattern.kt
```diff
@@ -50,7 +49,7 @@ object DateTimePattern : Pattern, ScalarType {
     override val typeAlias: String?
         get() = null
 
-    override val typeName: String = "datetime"
+    override val typeName: String = "date-time"
```
joelrosario: This change might break XML at some point which supports date and expects it to be datetime, and uses "($typeName)".

core/src/main/kotlin/io/specmatic/core/pattern/Discriminator.kt
```diff
@@ -87,7 +90,8 @@ class Discriminator(
             failure.breadCrumb(breadcrumb)
         }
 
-        return Result.Failure.fromFailures(failuresWithUpdatedBreadcrumbs)
+        val didNotMatchAnyOptionFailure = Failure(message = "Value didn't match any of the schema options", ruleViolation = StandardRuleViolation.ONE_OF_VALUE_MISMATCH)
```
joelrosario: Why do we need this extra error?

core/src/main/kotlin/io/specmatic/core/pattern/Discriminator.kt
```diff
@@ -164,32 +170,32 @@ class Discriminator(
...
         sampleData: Value?
-    ) = resolver.mismatchMessages.valueMismatchFailure("json object", sampleData)
+    ) = valueMismatchResult("json object", sampleData, resolver.mismatchMessages)
 
     private fun discriminatorKeyMissingFailure(discriminatorProperty: String, discriminatorCsv: String) =
         Result.Failure(
             "Discriminator property $discriminatorProperty is missing from the object (it's value should be $discriminatorCsv)",
             breadCrumb = discriminatorProperty,
-            failureReason = FailureReason.DiscriminatorMismatch
+            failureReason = FailureReason.DiscriminatorMismatch,
```
joelrosario, line 196: `failureReason` and `ruleViolation` appear very similar if you just look at the names. However, they serve different purposes, and maybe the names could make this clearer.

core/src/main/kotlin/io/specmatic/core/pattern/EmptyStringPattern.kt
```diff
@@ -2,15 +2,18 @@ package io.specmatic.core.pattern
...
+import io.specmatic.core.patternMismatchResult
 import io.specmatic.core.value.*
 
 object EmptyStringPattern : Pattern {
     override fun matches(sampleData: Value?, resolver: Resolver): Result {
         return when (sampleData) {
             EmptyString -> Result.Success()
-            else -> mismatchResult("empty string", sampleData, resolver.mismatchMessages)
+            is StringValue -> valueMismatchResult("empty string", sampleData, resolver.mismatchMessages)
+            else -> dataTypeMismatchResult("string", sampleData, resolver.mismatchMessages)
```
joelrosario: Should it be expected "empty string"?

core/src/main/kotlin/io/specmatic/core/pattern/QueryParameterScalarPattern.kt
```diff
@@ -17,23 +17,17 @@ data class QueryParameterScalarPattern(override val pattern: Pattern): Pattern b
...
 
     override fun matches(sampleData: Value?, resolver: Resolver): Result {
-        if(sampleData == null){
-            return Result.Failure("Null found, expected a scalar value")
-        }
+        if (sampleData == null) return dataTypeMismatchResult(pattern, sampleData, resolver.mismatchMessages)
         val sampleDataString = when (sampleData) {
             is ListValue -> {
-                if (sampleData.list.size > 1) return Result.Failure("Multiple values $sampleData found. Expected a single value")
+                if (sampleData.list.size > 1) return dataTypeMismatchResult(pattern, sampleData, resolver.mismatchMessages)
```
joelrosario: There's a specific error message that seems to be lost with this change about the count of query params.

core/src/main/kotlin/io/specmatic/core/pattern/QueryParameterScalarPattern.kt
```diff
@@ -17,23 +17,17 @@ data class QueryParameterScalarPattern(override val pattern: Pattern): Pattern b
...
         }
 
-        return try {
-            val parsedValue = runCatching { pattern.parse(sampleDataString, resolver) }.getOrDefault(StringValue(sampleDataString))
-            resolver.matchesPattern(null, pattern, parsedValue)
-        } catch (e: Throwable) {
-            Result.Failure(exceptionCauseMessage(e))
-        }
+        val parsedValue = runCatching { pattern.parse(sampleDataString, resolver) }.getOrDefault(StringValue(sampleDataString))
+        return resolver.matchesPattern(null, pattern, parsedValue)
```
joelrosario: is it possible for matchesPattern to throw an exception? If it does, it will bubble up as an exception, rather than a failure, since the try-catch has been removed.

core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
joelrosario: We can add the rule violation directly to the composite result.

core/src/main/kotlin/io/specmatic/core/pattern/AnyPattern.kt
joelrosario: Use datatype mismatch helper

core/src/main/kotlin/io/specmatic/core/pattern/DateTimePattern.kt
joelrosario: This should go back to "datetime"

core/src/main/kotlin/io/specmatic/core/pattern/Discriminator.kt
joelrosario: Same as anypattern

core/src/main/kotlin/io/specmatic/core/pattern/EmptyStringPattern.kt
joelrosario: Should still say "empty string" here.

core/src/main/kotlin/io/specmatic/core/pattern/QueryParameterScalarPattern.kt
joelrosario: Review the error message

core/src/test/kotlin/io/specmatic/core/BreadCrumbToJsonPathConverterTest.kt
```diff
@@ -0,0 +1,229 @@
+package io.specmatic.core
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
+import java.util.stream.Stream
+
+class BreadCrumbToJsonPathConverterTest {
```
joelrosario, line 10: From what I can see this class is only used in tests, in core. Is it used in Studio?

core/src/test/kotlin/io/specmatic/core/HttpQueryParamPatternTest.kt
```diff
@@ -443,33 +501,46 @@ class HttpQueryParamPatternTest {
             val stubbedNumericScalarQueryParameterPattern = HttpQueryParamPattern(mapOf("product_id" to QueryParameterScalarPattern(ExactValuePattern(NumberValue(1)))))
             val result = stubbedNumericScalarQueryParameterPattern.matches(HttpRequest("GET", "/", queryParams = QueryParameters(paramPairs = listOf("product_id" to "abc"))),  Resolver())
             assertThat(result is Failure).isTrue
-            assertThat(result.reportString().trimmedLinesList()).isEqualTo("""
-                >> PARAMETERS.QUERY.product_id
-
-                   Expected 1 (number), actual was "abc"
-            """.trimIndent().trimmedLinesList())
+            // TODO: This should've been TYPE_MISMATCH, the parse is throwing it off
```
joelrosario: Does this need attending to

core/src/test/kotlin/io/specmatic/core/HttpStubTests.kt
```diff
@@ -843,18 +876,50 @@ Scenario: JSON API to get account details with fact check
...
-            Expected number, actual was "DEF"
-            >> REQUEST.PARAMETERS.QUERY.creatorId
-            Expected number, actual was "ABC"
-            >> REQUEST.PARAMETERS.HEADER.PET-ID
-            Expected number, actual was "DEF"
-            >> REQUEST.BODY.creatorId
-            Expected number, actual was "ABC"
-            >> RESPONSE.BODY.petId
-            Expected number, actual was "DEF"
+            ${
```
joelrosario, line 877: The errors that show up seem confusing. Seems like 3 errors, but it's actually the same mismatch causing two errors..?

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityKtTest.kt
```diff
@@ -1253,7 +1253,7 @@ paths:
         val result: Results = testBackwardCompatibility(oldContract, newContract)
 
         assertThat(result.report()).contains("string in the new contract")
-        assertThat(result.report()).contains("number in the old contract")
+        assertThat(result.report()).contains("(number) in the old contract")
```
joelrosario: how come string has no brackets but "(number)" does?

core/src/main/kotlin/io/specmatic/core/pattern/EmptyStringPattern.kt
```diff
@@ -21,16 +24,16 @@ object EmptyStringPattern : Pattern {
...
-            else -> throw ContractException("""No data was expected, but got "$value" instead""")
+            else -> throw ContractException("No data was expected")
         }
     }
 
     override fun encompasses(otherPattern: Pattern, thisResolver: Resolver, otherResolver: Resolver, typeStack: TypeStack): Result {
-        if(otherPattern is EmptyStringPattern) return Result.Success()
-        return Result.Failure("No data was expected, but got \"${otherPattern.typeName}\" instead")
+        if (otherPattern is EmptyStringPattern) return Result.Success()
+        return patternMismatchResult(this, otherPattern, thisResolver.mismatchMessages)
```
joelrosario: "no data" was meant to represent in a backward compatibility check that no data was expected... it's a colloquial usage as there really isn't a type here

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityKtTest.kt
joelrosario: It should say X type in the old Y type in the new, like before this PR (how it works in main)

core/src/test/kotlin/io/specmatic/core/HttpStubTests.kt
joelrosario: PATH_MISMATCH should not be show in this case

PR #2156, opened 2025-12-29
Fix case-insensitive Host header removal in proxy

icholy:
**What**:

Fix case-insensitive Host header removal in proxy mode.

**Why**:

The `withoutDuplicateHostHeader` function was using `headers - "Host"` which only removes headers with exact case match. HTTP clients like node-fetch send lowercase `host:` headers, which were not being removed. This caused duplicate Host headers to be forwarded to the target server, resulting in 421 TLS certificate mismatch errors from CDNs like Fastly.

**How**:

Updated the Host header removal logic to use case-insensitive matching.

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Documentation added/updated (share link) N/A
- [ ] Sample Project added/updated (share link) N/A
- [ ] Demo video (share link) N/A
- [ ] Article on Website (share link) N/A
- [ ] Roadmpap updated (share link) N/A
- [ ] Conference Talk (share link) N/A

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/HttpRequest.kt
+++ b/core/src/main/kotlin/io/specmatic/core/HttpRequest.kt
@@ -289,7 +289,7 @@ data class HttpRequest(
             return headers
 
         if (isNotIPAddress(url.host))
-            return headers - "Host"
+            return headers.filterKeys { !it.equals("Host", ignoreCase = true) }
 
         return headers
     }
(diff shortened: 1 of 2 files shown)


joelrosario (approved):

joelrosario:
Thanks for the PR, @icholy!

PR #2189, opened 2026-01-19
feat: add support for regex matcher using pattern key with \$match

yogeshnikam671:


core/src/main/kotlin/io/specmatic/core/utilities/YAMLSerialisation.kt
```diff
@@ -14,7 +14,11 @@ val yamlMapper: ObjectMapper = ObjectMapper(yamlFactory).registerKotlinModule()
 
 fun yamlStringToValue(stringContent: String): Value {
     val cleaned = stringContent.cleanBOM()
-    val raw: Any = yamlMapper.readValue(cleaned, Any::class.java)
+    val raw: Any = try {
```
joelrosario: This modification result in a JSON `null` value being returned if there's a parse error. Need to better understand why that's happening.

yogeshnikam671: Thanks for pointing this out. I've resolved the same.

StarKhan6368 (approved):
Looks good to me

PR #2231, opened 2026-02-01
fix: fix the path in not covered test records for stub

yogeshnikam671:


core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
StarKhan6368: This doesn't need to change, does it ? And can the contentType be null ?

joelrosario (approved):

PR #2279, opened 2026-02-17
chore: update specmatic reporter version which now render actuator url in html report.

vedubhat:
Reason for change: 
This PR updates the specmatic reporter version to 0.3.1. This new reporter version now ensures that actuator url is rendered in html report if present in the specmatic configuration file.


yogeshnikam671 (approved):
LGTM

PR #2289, opened 2026-02-20
chore: update specmatic reporter version to support rendering of WIP coverage status in html reporter

vedubhat:
Reason to change: 
Update specmatic  reporter version to 0.3.3 which now supports rendering of WIP coverage status in remarks column in operation table in html reporter.


yogeshnikam671 (approved):
LGTM

PR #2337, opened 2026-03-12
fix: preserve original pattern in matchers for consistent matching and generation

yogeshnikam671:
Reason for change -
Matchers did not retain enough information about the original pattern when trying to derive a pattern from them.
With this change, each matcher derives a concrete pattern from its own kind using the original pattern, so the same pattern can be used consistently for both matching and generation of values.

Trigger for change -
RegexMatcher was supported on the test side, but it did not behave correctly on the mock side.
Investigation showed that the pattern being derived from RegexMatcher was incorrect, which caused mock-side handling to fail.


StarKhan6368 (approved):
Looks good to me

PR #2339, opened 2026-03-14
fix: return a not equals pattern from not equals matcher's patternFrom

yogeshnikam671:
Reason for change:
The pattern returned from not equals matcher was not able to handle enums properly where it would generate value which is not in the enum values set. This change ensures that not equals matcher always returns a pattern which will surely generate a value not equal to specified value while keeping the original pattern in tact.


yogeshnikam671:
Not Applicable for now. Closing it off.

PR #2343, opened 2026-03-17
Handle OpenAPI uri format as URLPattern in mocks

ketan:
~~This is however missing support for a `pattern` field to be able to generate specific type of URIs (for e.g. `tcp://...` or `jdbc://...` or `mqtt://...`)~~

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
@@ -2200,6 +2200,23 @@ class OpenApiSpecification(
         return StringPattern.from(stringConstraints, schema.pattern, example, collectorContext)
     }
 
+    private fun uriPattern(schema: Schema<*>, patternName: String, collectorContext: CollectorContext, example: String? = null): Pattern {
+        val baseUriPattern = URIPattern()
+        val regex = stringPattern(schema, patternName, collectorContext, example).regex ?: return baseUriPattern
+
+        return collectorContext.at("pattern").safely(
+            fallback = { baseUriPattern },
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -2200,6 +2200,23 @@ class OpenApiSpecification(
...
+        val baseUriPattern = URIPattern()
+        val regex = stringPattern(schema, patternName, collectorContext, example).regex ?: return baseUriPattern
+
+        return collectorContext.at("pattern").safely(
+            fallback = { baseUriPattern },
+            message = "URI regex constraint could not be applied, defaulting to uri"
+        ) {
+            RegexConstrainedPattern(
+                basePattern = baseUriPattern,
+                regex = regex,
```
joelrosario, line 2213: any reason we have support for regex, but not for minLength / maxLength? What if we embed the string pattern on line 2205 in the URIPattern, so that the string pattern can handle the other constraints and URIPattern handles only the uri format?

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -2200,6 +2200,23 @@ class OpenApiSpecification(
...
+                resolver = Resolver(),
+                eagerRegexValidation = false
```
joelrosario, line 2215: This should not be false

ketan:
Basis [Joel's comment](https://github.com/specmatic/specmatic/pull/2343#discussion_r2953866690) - it feels that the expectation is that of a wider scope. Closing this PR out.

PR #2344, opened 2026-03-17
chore: update patternFrom signature so that either deep pattern or exact pattern can be derived using the same method

yogeshnikam671:
Reason for change -
The existing patternFrom implementation always returns an exact value pattern if there is no matcher or pattern token involved. This is desirable on the mock side but on the test side during matching we'd want to match using the deep pattern and not the exact pattern.
This change parameterises patternFrom so that it can return both exact pattern and deep pattern based on our need.


StarKhan6368 (approved):
Looks good to me

PR #2346, opened 2026-03-17
fix: fix the logic in list pattern's pattern from so that it works as expected when deep pattern is expected from an example

yogeshnikam671:
Reason for change -
The deep pattern derived from an example containing a list is currently mandating that the length of the array should be same as that of the value in the example. This is not expected behaviour. This PR fixes it so that the appropriate deep pattern is computed.


yogeshnikam671:
This needs more discussion

PR #2347, opened 2026-03-17
chore: use http request matcher executor to match requests in http stub

yogeshnikam671:


core/src/main/kotlin/io/specmatic/core/pattern/Pattern.kt
```diff
@@ -119,7 +119,8 @@ fun patternFromValueUsing(
     if(value !is StringValue) return parseValueToType(value)
     if(isPatternToken(value)) return DeferredPattern(value.string)
     if(isMatcherToken(value)) {
-        return Matcher.patternFrom(
+        val matcherEngine = MatcherEngine.loadOrThrow()
```
StarKhan6368: Should we throw an error here instead of defaulting to the originalPattern ? cc @joelrosario

core/src/main/resources/META-INF/services/io.specmatic.core.MatcherEngine
```diff
@@ -0,0 +1 @@
+io.specmatic.core.matchers.DefaultMatcherEngine
```
StarKhan6368: This doesn't exist ? Should we be implementing or removing the registration ?

core/src/main/kotlin/io/specmatic/test/ScenarioAsTest.kt
```diff
@@ -133,19 +134,21 @@ data class ScenarioAsTest(
...
-                )
-                if(matchesResult is Result.Failure) {
-                    return ContractTestExecutionResult(
-                        result = matchesResult.withBindings(testScenario.bindings, response),
-                        request = request,
-                        response = response
+
+            if(responseBodyFromExample != null) {
+                matcherEngine?.let {
+                    val matchesResult = it.matchesResult(
```
StarKhan6368, line 140: Not part of this PR, but just to confirm, we do not support matchers in headers ?

StarKhan6368, line 140: Also to confirm, are the fixtures are execute pre-contract test validation ? is that intended ?

StarKhan6368 (changes requested):

core/src/main/resources/META-INF/services/io.specmatic.core.MatcherEngine
yogeshnikam671: Good catch. This file is not supposed to exist. Removed the same.

core/src/main/kotlin/io/specmatic/core/pattern/Pattern.kt
yogeshnikam671: IMO, throwing an error makes it explicit that matcher tokens are not supported.

core/src/main/kotlin/io/specmatic/test/ScenarioAsTest.kt
yogeshnikam671: At the moment, we do not. We'll enhance this when needed. I think by headers, you mean response headers?

yogeshnikam671: Good catch. Addressed.

core/src/main/kotlin/io/specmatic/test/ScenarioAsTest.kt
```diff
@@ -206,6 +198,16 @@ data class ScenarioAsTest(
 
             testScenario.exampleRow?.let { ExampleProcessor.store(it, request, responseToCheckAndStore) }
 
+            if(result !is Result.Failure) {
+                val afterFixtureExecutionResult = fixtureExecutionResult(AFTER_FIXTURE_DISCRIMINATOR_KEY)
+                if (afterFixtureExecutionResult.isSuccess().not()) {
+                    return ContractTestExecutionResult(
+                        result = afterFixtureExecutionResult,
```
StarKhan6368: We should bind the result with the scenario and response, as we do with the others, same for beforeFixture return at line 126

StarKhan6368 (changes requested):

core/src/main/kotlin/io/specmatic/test/ScenarioAsTest.kt
yogeshnikam671: Addressed

StarKhan6368 (approved):
Looks good to me

PR #2348, opened 2026-03-18
chore(deps): Remove mozilla-rhino "dependency"

ketan:
This was upgraded to work around some vulnerability issues that do not exist as of today.


yogeshnikam671 (approved):
LGTM

PR #2358, opened 2026-03-23
OpenAPI Spec Conformance Tests: Part 1

sahilm:
**What**:

Introduce OpenAPI Spec conformance tests. In part 1 we validate and document that Specmatic supports `001-http-methods` and `002-path-parameters` features from the OpenAPI specification (version 3.0.x) by running a loop test. For example a loop test of the specification `002-path-parameters/007-two-params-with-separator.yaml` starts a Specmatic mock for this specification and then points a Specmatic test at it. A successful test validates that Specmatic can load, parse, mock and test such a specification. 

**Why**:

This ensures and documents that Specmatic is and remains compliant with the OpenAPI specification.

**Design**:

1. All of Specmatic is the System Under Test and is treated as a block box. Only the public interface (exit code and request/responses) are used to validate Specmatic's behaviour.
2. Since these are long running black box tests it's necessary to run them in parallel. We intend to add several hundreds of tests to this suite in follow up PRs. The tests depend on the docker image of Specmatic only.
3. The tests live in an independent gradle project. To run the tests please run `./gradlew :conformance-tests:check -PenableConformanceTests=true`. These tests will be wired up to the main build in subsequent PRs.

**Follow up PRs**:
1. Wire up in the main build and documentation.
1. Detailed request/response validation.
1. Header and Content-Type validations.
1. More features of OpenAPI will be exercised.

- [X] Unit Tests
- [X] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)

### DIFF
--- a/conformance-tests/build.gradle.kts
+++ b/conformance-tests/build.gradle.kts
@@ -29,3 +29,57 @@ if (enableConformanceTests?.toBoolean() == true) {
         enabled = false
     }
 }
+
+val generateConformanceTests by tasks.registering {
+    val specsDir = file("src/test/resources/specs")
+    val outputDir = file("build/generated/sources/conformance/kotlin")
+
+    inputs.dir(specsDir)
+    outputs.dir(outputDir)
+
+    doLast {
+        outputDir.deleteRecursively()
+        outputDir.mkdirs()
+
+        val specFiles = specsDir.walkTopDown()
+            .filter { it.isFile && it.extension in listOf("yaml", "yml") }
+            .map { it.relativeTo(specsDir).path.replace("\\", "/") }
+            .sorted()
+            .toList()
+
+        val header = "package conformance_tests\nimport org.junit.jupiter.api.DisplayName\n"
+
+        val classes = specFiles.joinToString("\n") { relativePath ->
+            val segments = relativePath.split("/")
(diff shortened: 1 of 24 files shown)


conformance-tests/src/test/kotlin/conformance_tests/AbstractConformanceTest.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val openAPISpecFile: String
+) {
+    @Test
+    fun `loop test`() {
+        val dockerCompose = DockerCompose(
+            specmaticVersion = System.getProperty("specmatic.version"),
+            pathToOpenAPISpecFile = openAPISpecFile
+        )
+
+        try {
```
sahilm, line 16: Not using BeforeAll/AfterAll here because this will change soon in a follow up PR.

conformance-tests/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+        useJUnitPlatform()
+        systemProperty("specmatic.version", project.version)
+    }
+} else {
+    tasks.test {
+        enabled = false
+    }
+}
+
+val generateConformanceTests by tasks.registering {
```
sahilm, line 33: Need to use Codegen here because jUnit can only parallelize if each test lives in its own class.

PR #2359, opened 2026-03-23
chore: change output to outputs in TestResultRecord so that it works with latest reporter

yogeshnikam671:
Reason for change -
The reporter is updated to support multiple outputs for a singular input - https://github.com/specmatic/specmatic-reporter/pull/107
This PR makes the necessary changes to fix the compilation issues caused by latest reporter changes.


StarKhan6368 (approved):
Looks good to me

PR #2360, opened 2026-03-23
OpenAPI Spec Conformance Tests: Part 2

sahilm:
This is part 2 of https://github.com/specmatic/specmatic/pull/2358. See main PR for details. In this PR:
1. We validate that all operations (HTTP method + /path) defined in the specification were exercised in the tests and that no unspecified operations were performed.
2. We validate that request and response bodies are valid against the spec.


conformance-tests/src/test/kotlin/io/specmatic/conformance_tests/AbstractConformanceTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    private lateinit var httpExchanges: List<HttpExchange>
+
+    private val spec: OpenApiSpec =
+        OpenApiSpec(File("${workDir.absolutePath}/${specsDirName}/$openAPISpecFile"))
+
+
+    @BeforeAll
+    fun setup() {
+        dockerCompose = DockerCompose(
+            specmaticVersion = System.getProperty("specmatic.version") ?: "latest",
```
ketan: Use the automated `VersionInfo` class

sahilm:
Addressed feedback @ketan:
>Possible gap where a bug in specmatic generates an empty request and/or response when it should have. We need a stricter check to validate conformance from spec -> traffic and traffic back to spec.

We now convert all httpExchanges -> Operation and compare that the set of operations defined in the spec is equal to the set of operations derived from all the httpExchanges.

>Possibly make this more intent revealing

Same as above. Hopefully it's better now.

>Operation consists of PATH, Method, Status Code, Request Content type (as things stand currently)

Same as above.

PR #2365, opened 2026-03-24
Issue 1741: Include min/max Items on ListPattern

### LINKED ISSUE #1741: minItems/maxItems in array
Hi team, 
I am able to see that usage of minItems in an array schema validation is not wokrking. Can you confirm if we support this or not

Also I am not able to add an assertion that if i am getting a list ,it must not be empty. I thing this overlps with the above question only

EDIT:
I came across other use case where I wanted to make sure that the returning list must contain a string named "ABC" . I was not able to get through this. Is there some way we can do this?

@joelrosario

abhiverma31:
**What**: Added minItems and maxItems validation support for array schemas in ListPattern.

**Why**: Users were unable to enforce array length constraints (minItems/maxItems) defined in their OpenAPI specs. Specmatic is parsing but ignoring these constraints, causing validation to pass for arrays that violated the specified limits.

**How**:
1. Added `minItems` and `maxItems` nullable Int properties to `ListPattern` data class
2. Implemented validation logic in `ListPattern.matches()` to check array length against these constraints
3. Added unit test to verify the validation

**Checklist**:

- [ ] Unit Tests
- [ ] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)

**Issue ID**:
Closes: #1741

This is for an initial review. If the approach is fine, we'd proceed with more testing.


yogeshnikam671:
Hi @abhiverma31, 

Thanks for noticing this bug and raising the PR. The changes you’ve made are in the right direction, but some more wiring is needed to take this to completion.

A little context on how patterns are used in Specmatic:
A pattern has two major responsibilities: `matching values` and `generating values`.

The matching part is used:
- on the mock/stub side, to match the incoming request
- on the test side, to match the response from the system under test

The generation part is used:
- on the mock/stub side, to generate a response for the incoming request
- on the test side, to generate a request that will be sent to the system under test

The part you’ve addressed in the pattern is the matching. Generation is not addressed yet.

When generating a value from a list pattern with `minItems` and `maxItems` constraints, the generated list should have a size that is:
- greater than or equal to `minItems`
- less than or equal to `maxItems`

Apart from this, you’ll also need to make sure these constraints are parsed when the spec is read by `OpenApiSpecification`, and that they are set correctly while creating `ListPattern`.

Please also add an end-to-end test for this behaviour. A good way to do that is:
- start a stub using an OpenAPI spec that contains these constraints
- run tests against that stub

This should cover:
- matching on the stub side
- matching on the test side
- generation on the stub side
- generation on the test side

You can refer to this test for the general structure:
`should be able to load and respond to requests using xml based openAPI Specification with external example`

Take a stab at these first. There are a few more things that still need to be taken care of, which I’ll call out once the above is addressed.

abhiverma31:
Thank you @yogeshnikam671 for your detailed feedback. I've tried to incorporate the review comments based on your feedback, please review it and suggest further changes. 

- extract minItems and maxItems from OpenAPI specs
- generate random arrays with sizes that respect minItems/maxItems constraints
- integration test as suggested

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -106,6 +108,14 @@ data class ListPattern(
                 else -> dataTypeMismatchResult(this, sampleData, resolver.mismatchMessages)
             }
 
+        if (minItems != null && sampleData.list.size < minItems) {
+            return Result.Failure("Expected at least $minItems items, got ${sampleData.list.size}")
```
yogeshnikam671: Update the message to indicate that it is an issue with list with something like "List is expected to contain at least $minItems items, but it contained ${sampleData.list.size} items".
Also set the ruleViolation as StandardRuleViolation.CONSTRAINT_VIOLATION.
You can refer to line 122 for this.

yogeshnikam671: Similarly update the `maxItems` specific failure.

core/src/test/kotlin/io/specmatic/core/pattern/ListPatternTest.kt
```diff
@@ -981,4 +981,31 @@ Feature: Recursive test
             assertThat(paths).allMatch { it.matches(Regex("\\{\\[\\d+]\\}\\{string\\}")) }
         }
     }
+
+    @Test
+    fun `should reject array with fewer items than minItems`() {
```
yogeshnikam671, line 963: Add test where the array has items more than maxItems too.

core/src/test/kotlin/io/specmatic/stub/HttpStubKtTest.kt
```diff
@@ -1843,4 +1844,27 @@ paths:
         assertThat(results.success()).withFailMessage(results.report()).isTrue()
         assertThat(results.testCount).isEqualTo(1)
     }
+
+    @Test
+    fun `should respect minItems and maxItems constraints in stub generation and validation`() {
```
yogeshnikam671, line 1849: Add a similar test where the request contains an array in the request body with the minItems/maxItems constraints.

core/src/main/kotlin/io/specmatic/core/Resolver.kt
```diff
@@ -336,12 +337,18 @@ data class Resolver(
     fun generateList(pattern: ListPattern): Value {
```
yogeshnikam671, line 337: Write tests for this method.

yogeshnikam671 (changes requested):
Hi @abhiverma31, sorry for responding late. The changes look good. You've addressed almost all the feedbacks.
I've left some more feedbacks on specific code blocks below.

Also, now you will have to look for all the instantiations of ListPattern across the repo and make sure you set the minItems, maxItems wherever applicable at all those places.

Let me know if you have any further questions, I'll be happy to help out.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: Done

core/src/test/kotlin/io/specmatic/core/pattern/ListPatternTest.kt
abhiverma31: Done

core/src/test/kotlin/io/specmatic/stub/HttpStubKtTest.kt
abhiverma31: Done

core/src/main/kotlin/io/specmatic/core/Resolver.kt
abhiverma31: Done

abhiverma31:
@yogeshnikam671 Thanks again for your time on reviewing this. I've made the changes as suggested:
 - updated error messages
 - added maxItems validation test 
 - added request body validation test 
 - added Resolver.generateList() tests

On checking the various instantiations of ListPattern(), I thought the only place required is OpenApiSpecification.kt - which I've done. I was not sure if I need to modify the ListPattern instantiations in Grammar.kt and Feature.kt - and whether there is any negative impact on not including the min/max fields ~ going by the tests are succeeding.
The rest are majority test classes. 

Please advise if my assessment is incorrect.

core/src/test/kotlin/io/specmatic/stub/HttpStubKtTest.kt
```diff
@@ -1843,4 +1844,41 @@ paths:
...
+                    return response
+                }
+            })
+        }
+
+        assertThat(results.success()).withFailMessage(results.report()).isTrue()
+    }
+
+    @Test
+    fun `should validate minItems and maxItems constraints in request body`() {
```
yogeshnikam671, line 1873: This test is good. But we need a test where we do `scenario.executeTests` and for each request generated by test which is received in the execute method should be asserted on with the assertion that the request contains the array within minItems, maxItems bound.

yogeshnikam671 (changes requested):

yogeshnikam671:
Hey @abhiverma31, thanks for jumping on this so quickly. The changes look really good.

Just a couple of things:

You don’t need to set `minItems` / `maxItems` everywhere `ListPattern` is instantiated. Only do it at places where it is possible and makes sense, like ListPattern's `newBasedOn` and `negativeBasedOn`.

Also, one more thing about `negativeBasedOn`. Just adding some context here, this method is used to generate negative patterns, which are then used for generating negative tests.

Now that `ListPattern` has `minItems` and `maxItems`, we should also generate `ListPattern`s where:
* `minItems` is greater than the original `maxItems`, and `maxItems` becomes equal to this new `minItems`
* `maxItems` is less than the original `minItems`, and `minItems` becomes equal to this new `maxItems`

Do this only where it makes sense though i.e. only when it is possible to generate a list whose size falls within the resulting `minItems` and `maxItems`.

For example, say the original pattern has:

* `minItems = 0`
* `maxItems = 3`

If we try to generate a negative pattern by making `maxItems` less than the original `minItems`, we might derive:

* `newMaxItems = minItems - 1 = -1`

So we end up with:

* `maxItems = -1`
* `minItems = -1`

This doesn’t make sense because list size cannot be negative. So we should avoid generating such cases and only create these patterns when the derived values are valid.

Also, please add tests to cover these scenarios when you make the change.

core/src/test/kotlin/io/specmatic/stub/HttpStubKtTest.kt
abhiverma31: Ok implemented a test using `executeTests()` so that we are able to test an end-to-end integration.

abhiverma31:
@yogeshnikam671 Appreciate your continued support in reviewing the work.

I've made the the following changes, based on your recent feedback :
- added test for `negativeBasedOn` generating size violations
- added test for `newBasedOn` preserving constraints
- added test for POST request generation with executeTests verifying array bounds

I've ensured that minItems/maxItems constraints are properly handled in both `negativeBasedOn` (for generating constraint violations) and `newBasedOn` (for preserving constraints during pattern transformation). Added corresponding tests to verify these behaviors.

Also, as suggested - included an end-to-end test that validates the complete request generation flow.

Please me let me know if the changes meet all expectations, and if anything else is required to be done. Thank you.

abhiverma31:
Hi @yogeshnikam671  - just checking if you got a chance to review this PR. Please let me know if any further changes required on this. Thank you.

yogeshnikam671:
Hey @abhiverma31, sorry did not get a chance to review this. I'll update you once I do the same.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -173,12 +185,34 @@ data class ListPattern(
...
+                yieldAll(pattern.negativeBasedOn(row.stepDownIntoList(), resolver, config)
+                    .map { negativePatternValue ->
+                        negativePatternValue.ifValue { pattern ->
+                            ListPattern(pattern, minItems = minItems, maxItems = maxItems) as Pattern
+                        }.breadCrumb(LIST_BREAD_CRUMB)
+                    })
+
+                if (minItems != null) {
+                    val pattern = copy(
+                        minItems = minItems - 1,
```
yogeshnikam671: If minItems is 0, this creates a pattern with an invalid minItems & maxItems value i.e. -1. Add a test for this.
Also see if it is possible to do this without using the yield and yield all operators

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -3,14 +3,18 @@ package io.specmatic.core.pattern
...
 
 data class ListPattern(
     override val pattern: Pattern,
     override val typeAlias: String? = null,
     override val example: List<String?>? = null,
-    override val extensions: Map<String, Any>  = emptyMap()
+    override val extensions: Map<String, Any>  = emptyMap(),
+    val minItems: Int? = null,
+    val maxItems: Int? = null
 ) : Pattern, SequenceType, HasDefaultExample, PossibleJsonObjectPatternContainer {
```
yogeshnikam671, line 16: In an init block, let's add a validation that minItems and maxItems are valid non-negative integers. If that is not the case, let's throw a ContractException. Refer to StringPattern for a similar handling. And add a test for this.

core/src/test/kotlin/io/specmatic/core/pattern/ListPatternTest.kt
```diff
@@ -981,4 +982,80 @@ Feature: Recursive test
...
+            minItems = 2,
+            maxItems = 5
+        )
+
+        val negatives = pattern.negativeBasedOn(Row(), Resolver(), NegativePatternConfiguration())
+            .map { it.value }
+            .filterIsInstance<ListPattern>()
+            .toList()
+
+        val minItemsViolation = negatives.find { it.maxItems != null && it.maxItems < 2 }
```
yogeshnikam671: Do a negative.count {} of such and maxItemsViolation patterns.
Those counts should be non-zero and the total of those counts should be equal to negative.size

This should be the assertion here.

yogeshnikam671 (changes requested):

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: I've added a check for minItems. But not yet changed yield/yieldAll operators. I'll think how that can be done, my understanding is instead of creating sequences lazily, the suggestion is to create lists instead.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: I've mimicked how it's been done in StringPattern.

core/src/test/kotlin/io/specmatic/core/pattern/ListPatternTest.kt
abhiverma31: sure, instead of find, count is a more rigourous check. have added the asserts.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
yogeshnikam671: The check looks good. No, my suggestion is not to create lists but to have a single return statement for patterns. If yield is used, it becomes hard to follow all the returns of the method.

Here's a suggestion for you. If this works and passes all tests, you can use the following -
```kotlin
private fun listPatternsWithNegativePatternsWithin(
    row: Row,
    resolver: Resolver,
    config: NegativePatternConfiguration
): Sequence<ReturnValue<Pattern>> =
    pattern.negativeBasedOn(row.stepDownIntoList(), resolver, config)
        .map { negativePatternValue ->
            negativePatternValue.ifValue { pattern ->
                ListPattern(pattern, minItems = minItems, maxItems = maxItems) as Pattern
            }.breadCrumb(LIST_BREAD_CRUMB)
        }

override fun negativeBasedOn(
    row: Row,
    resolver: Resolver,
    config: NegativePatternConfiguration
): Sequence<ReturnValue<Pattern>> {
    return attempt(breadCrumb = LIST_BREAD_CRUMB) {
        val minItemsViolationPattern = minItems
            ?.takeIf { it > 0 }
            ?.let {
                HasValue(
                    copy(minItems = it - 1, maxItems = it - 1),
                    "is set to a value with items less than minItems '$it'"
                )
            }

        val maxItemsViolationPattern = maxItems
            ?.let {
                HasValue(
                    copy(minItems = it + 1, maxItems = it + 1),
                    "is set to a value with items greater than maxItems '$it'"
                )
            }

        listPatternsWithNegativePatternsWithin(row, resolver, config) +
            listOfNotNull(minItemsViolationPattern, maxItemsViolationPattern).asSequence()
    }
}
```

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
yogeshnikam671: This is good. But I'd prefer not have the variable `effectiveMinItems`.
We could simply add the null checks are places where we want to operate on or access `minItems`. Let's get rid of this extra variable `effectiveMinItems`

yogeshnikam671 (changes requested):

core/src/test/kotlin/io/specmatic/core/pattern/ListPatternTest.kt
yogeshnikam671: Thanks, the changes look good.

abhiverma31:
@yogeshnikam671 I've addressed both the comments:

On the first one - re: not using `yield` constructs, I like your suggestion of creating a separate helper function for the inner pattern type check. I just had one suggestion on the same lines - we can even separate out the check for size constraints in its own helper function. So that `negativeBasedOn()` is just aggregating the results from the two helper functions. In my view, this way we have better uniformity/ readability. I've implemented it this way and pushed my code. Let me know if you agree, happy to know your thoughts. 

On the second one - I've got rid of `effectiveMinItems` as suggested and doing an in place null check instead. I just had to make a corresponding change in the message of one of the tests that is checking for maxItems > 0, it wasn't too reflective of what we are testing ..earlier.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -10,8 +10,27 @@ data class ListPattern(
...
+    
+    init {
+        minItems?.let {
+            if (it < 0) {
+                throw ContractException("minItems $it cannot be less than 0")
```
yogeshnikam671: "minItems must be >= 0, but got $it"
This reads better is what I feel. I am okay with anything else as well. The current message that we have will not read good.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -10,8 +10,27 @@ data class ListPattern(
...
+            }
+        }
+        maxItems?.let {
+            if (it < 0) {
+                throw ContractException("maxItems $it cannot be less than 0")
```
yogeshnikam671: "maxItems must be >= 0, but got $it"

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
```diff
@@ -10,8 +10,27 @@ data class ListPattern(
...
+            }
+        }
+        if (minItems != null && maxItems != null && minItems > maxItems) {
+            throw ContractException("maxItems $maxItems cannot be less than minItems $minItems")
```
yogeshnikam671: "maxItems ($maxItems) cannot be less than minItems ($minItems)"

yogeshnikam671 (changes requested):

yogeshnikam671:
Hey @abhiverma31, the changes look really good. Thanks for being so proactive in addressing the feedbacks. 
I think we are very close to merging this PR.
Once you fix the exception message specific feedback, I'll do some final sanity checks at my end.

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: Done

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: Done

core/src/main/kotlin/io/specmatic/core/pattern/ListPattern.kt
abhiverma31: Done

abhiverma31:
@yogeshnikam671 , agreed. Those messages make for better readability. Have pushed the changes.

yogeshnikam671 (approved):
LGTM

PR #2367, opened 2026-03-25
feat: add support to store and replace variables across before and after fixtures using generic ExecutionVariableContext

yogeshnikam671:
Reason for change -
Add ability to store and replace variables across the fixtures.
With this change we will be able to capture the response and access the fields from the response using the substitution token syntax e.g. ("$(FIXTURE.BEFORE.1.RESPONSE.id)")

### DIFF
--- a/core/src/main/kotlin/io/specmatic/test/variables/ExecutionVariableContext.kt
+++ b/core/src/main/kotlin/io/specmatic/test/variables/ExecutionVariableContext.kt
@@ -1,0 +1,90 @@
+package io.specmatic.test.variables
+
+import io.specmatic.core.pattern.ContractException
+import io.specmatic.core.value.JSONArrayValue
+import io.specmatic.core.value.JSONObjectValue
+import io.specmatic.core.value.StringValue
+import io.specmatic.core.value.Value
+import io.specmatic.test.ExampleProcessor.toFactStore
+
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/io/specmatic/test/variables/ExecutionVariableContext.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        val resolved = INLINE_SUBSTITUTION_PATTERN.replace(value) { match ->
+            getValue(match.groupValues[1]).toStringLiteral()
+        }
+
+        return convertInterpolatedString(resolved)
+    }
+
+    private fun lookupExactValue(value: String): Value? =
+        EXACT_SUBSTITUTION_PATTERN.matchEntire(value)?.groupValues?.get(1)?.let(::getValue)
```
StarKhan6368, line 89: This will fail for strings that start and end with substitution pattern, for example `$(FIXTURE.BEFORE.1.RESPONSE.id)-$(FIXTURE.BEFORE.1.RESPONSE.name)`

StarKhan6368 (changes requested):

yogeshnikam671:
Irrelevant. Needs more thinking.

PR #2375, opened 2026-03-26
feat: add ability to hook a before each hook per test

yogeshnikam671:
Reason for change -
We needed an ability to run a before each hook prior to any kind of test (be it openapi/grpc/async, etc.)


StarKhan6368 (approved):
Looks good to me

PR #2387, opened 2026-03-30
Include error output in Docker commands and reduce default concurrency levels

sahilm:
This helps with debugging tests. No major impact on total test run time after reducing concurrency down to a fixed level of 4.


SaachiKaup (approved):
Approved

PR #2409, opened 2026-04-10
fix (test-result-record): handle test result message for all test result types

vedubhat:
Reason for change -
The HTML report was missing the details around why a certain test failed especially when it comes to a "missing in spec" test.
This caused because the ctrf report which is used to render the html report, itself did not have those details.
This PR makes sure that the error details are set in the ctrf report during its generation.
This is done by relying directly on the test result while setting these details in the ctrf report.


yogeshnikam671 (commented):
LGTM

PR #2414, opened 2026-04-11
Ensure positive mutations from OptionalBody include relevant details

StarKhan6368:
**What**: Ensure positive mutations from OptionalBody include relevant details

**Why**: The ValueDetails should be preserved when executing newBasedOn, as they provide the user with information about positive mutations

**How**: Fix in `AnyPattern.newBasedOn` that causes `ValueDetails` to be omitted during pattern iteration

**Checklist**:

- [x] Unit Tests
- [ ] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/NoBodyPattern.kt
+++ b/core/src/main/kotlin/io/specmatic/core/NoBodyPattern.kt
@@ -16,7 +16,7 @@ object NoBodyPattern : Pattern {
 
     override fun generate(resolver: Resolver): Value = NoBodyValue
 
-    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(this, "has been omitted"))
+    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(this))
 
     override fun newBasedOn(resolver: Resolver): Sequence<Pattern> = sequenceOf(this)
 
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/io/specmatic/core/NoBodyPattern.kt
```diff
@@ -16,7 +16,7 @@ object NoBodyPattern : Pattern {
 
     override fun generate(resolver: Resolver): Value = NoBodyValue
 
-    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(this))
+    override fun newBasedOn(row: Row, resolver: Resolver): Sequence<ReturnValue<Pattern>> = sequenceOf(HasValue(this, "has been omitted"))
```
joelrosario: Ideally this would appear when generating negative tests of a body, which is in HttpRequestPattern.negativeBasedOn. It's already there. NoBodyPattern is a general use class, so the fact that this method is called may not automatically mean that the request body was dropped.

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/NoBodyPattern.kt
StarKhan6368: We still need it for `newBasedOn` for `OptionalBodyPattern`, We should explain the reasoning for a positive generated test case that happens when the requestBody is optional. It might be better to override the functionality in `OptionalBodyPattern` and add the reasoning there ?

PR #2421, opened 2026-04-16
Skip asyncapi files during OpenAPI bcc

sahilm:
The current design conflates the collection of changes to specs and examples. This means that we must filter out asyncapi files with a negative check when loading up all files.

### DIFF
--- a/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
+++ b/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
@@ -28,9 +28,14 @@ class BackwardCompatibilityCheckCommandV2 : BackwardCompatibilityCheckBaseComman
         return testBackwardCompatibility(oldFeature as Feature, newFeature as Feature)
     }
 
+    companion object {
+        private val jsonMapper = ObjectMapper()
+        private val yamlMapper = ObjectMapper(YAMLFactory())
+    }
+
     private fun isYAML(string: String): Boolean {
         return try {
-            ObjectMapper(YAMLFactory()).readValue(string, Map::class.java)
+            yamlMapper.readValue(string, Map::class.java)
             true
         } catch (e: Throwable) {
             false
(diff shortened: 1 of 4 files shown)


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -39,18 +44,33 @@ class BackwardCompatibilityCheckCommandV2 : BackwardCompatibilityCheckBaseComman
...
+    private fun isAsyncAPI(string: String): Boolean {
+        return try {
+            val map = yamlMapper.readValue(string, Map::class.java)
+            map.containsKey("asyncapi")
+        } catch (e: Throwable) {
+            false
+        }
+    }
+
     override fun File.isValidFileFormat(): Boolean {
```
yogeshnikam671, line 63: I think we can simply reuse `isValidSpec` check here.
The downside of that could be that if there is a genuine openapi spec which is malformed, it won't be picked up by the BCC.
I think we could handle this differently where maybe at the end of the bcc logs, we report all changed files that were skipped.
That way, the user knows that the malformed openapi spec is skipped.

Having an asyncapi specific check in openapi's bcc command seems a little out of place. Otherwise functionally this looks good.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -227,10 +227,14 @@ class OpenApiSpecification(
         }
 
         fun getParsedOpenApi(openApiFilePath: String): OpenAPI {
+            // TODO: This is ignoring parse errors for example it declares asyncapi files as valid and parseable
```
yogeshnikam671, line 230: What I understood from this is that `OpenApiParser.read`  is parsing asyncapi files as valid specs. So we should update the check by checking the returned result for no messages.
We can discuss this.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
yogeshnikam671: As discussed, we will tackle this later as it is a fundamental issue with the bcc framework itself.

yogeshnikam671 (approved):
LGTM

PR #2422, opened 2026-04-17
Make `fixValue` in ListPattern respect the minItems and maxItems constraints

yogeshnikam671:
Recently, support for `minItems` and `maxItems` constraints was added in `ListPattern` but unfortunately we missed to cover `fixValue` in that change.

This PR fixes that by making sure `fixValue` always returns a list of valid size as per the specified constraints.
If no constraints are specified, the existing behaviour kicks in.

See the tests added in `ListPatternTest` to understand what all cases are covered.


yogeshnikam671:
Deferred to later because of https://github.com/specmatic/specmatic/pull/2424.
Closing out.

PR #2424, opened 2026-04-17
Revert "Issue 1741: Include min/max Items on ListPattern"

joelrosario:
Reverts specmatic/specmatic#2365

There are additional features to be implemented as per feedback from the team.


yogeshnikam671:
Hi @abhiverma31, just wanted to update you that we’ll be reverting the minItems/maxItems changes from the ListPattern PR for now.

There’s nothing wrong with the implementation itself, it works well. However, I missed an important aspect that was later identified by @joelrosario when this feature was used in a real scenario.

The issue is that when minItems or maxItems have large values, and the list elements are themselves large objects, Specmatic ends up generating very large payloads during tests. This becomes a practical problem for customers who use bigger constraint values.

We’re currently working internally on a cleaner and more scalable way to handle such cases across the board. Once that approach is ready, we’ll bring these changes back.

There are also a few missing pieces in the current implementation. I’ll raise issues for those separately so you can refer to them and address them if you'd like.

Apologies for the inconvenience. I know you’ve put significant effort into this, and we truly appreciate it. We need to make this change for now to unblock existing customers facing issues with larger constraints.

I’ll keep you posted once we re-merge the reverted PR and bring the feature back. Thanks.

abhiverma31:
@yogeshnikam671 Thanks for the update. Sure, please do tag me here when you've merged the PR back. And any follow up work that emerges as a result, I'd be happy to look at them.

PR #2440, opened 2026-04-22
Skip status in failure reports if status == 0

sahilm:
If the status field is not initialized or unset we want to skip it. This is because we now have ScenarioDetails for AsyncAPI specs where status codes aren't a meaningful concept.


yogeshnikam671 (approved):
LGTM

PR #2444, opened 2026-04-23
Allow configuration of failure report headings.

sahilm:
Allow configuration of failure report headings.


sahilm:
Paired with @yogeshnikam671 . Merging.

PR #2449, opened 2026-04-24
Fix distinctReport to truly de-duplicate the failures

yogeshnikam671:
Fix `distinctReport` to truly de-duplicate the error messages by grouping failure reports per scenario and merging the distinct failures

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/FailureReport.kt
+++ b/core/src/main/kotlin/io/specmatic/core/FailureReport.kt
@@ -1,6 +1,18 @@
 package io.specmatic.core
 
 data class FailureReport(val contractPath: String?, private val scenarioMessage: String?, val scenario: ScenarioDetailsForResult?, private val matchFailureDetailList: List<MatchFailureDetails>): Report {
+
+    fun groupingKey(): String {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/io/specmatic/core/FailureReport.kt
```diff
@@ -1,6 +1,18 @@
 package io.specmatic.core
 
 data class FailureReport(val contractPath: String?, private val scenarioMessage: String?, val scenario: ScenarioDetailsForResult?, private val matchFailureDetailList: List<MatchFailureDetails>): Report {
+
+    fun groupingKey(): String {
+        if(contractPath != null && scenario != null) return "$contractPath ${scenario.name}"
```
joelrosario: We're depending on scenario.name to be unique. Ideally it should be but is it guaranteed?

joelrosario (changes requested):

core/src/main/kotlin/io/specmatic/core/FailureReport.kt
joelrosario: This was handled by using operation tokens in the grouping key.

joelrosario (approved):

PR #2484, opened 2026-05-07
chore: pass before and after fixture execution details to CtrfTestMetadata in ctrf report

vedubhat:
Reason to change: 
Pass fixture execution results to the CTRF report so they can be rendered in the HTML reporter, keeping the html-reporter output in sync with Studio, which already displays fixture execution results.


yogeshnikam671 (approved):
LGTM, merge when tests go green.

PR #2486, opened 2026-05-07
Improve backward compatibility check performance

StarKhan6368:
**What**: Improves backward compatibility check performance by reducing unnecessary generation and matching work in hot paths

**Checklist**:
- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate N/A
- [ ] Security scans don't report any vulnerabilities N/A
- [ ] Documentation added/updated (share link) N/A
- [ ] Sample Project added/updated (share link) N/A
- [ ] Demo video (share link) N/A
- [ ] Article on Website (share link) N/A
- [ ] Roadmpap updated (share link) N/A
- [ ] Conference Talk (share link) N/A

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/Resolver.kt
+++ b/core/src/main/kotlin/io/specmatic/core/Resolver.kt
@@ -459,7 +459,7 @@ data class Resolver(
             return null
         }
 
-        val path = dictionaryLookupPath.replace(WILDCARD_INDEX, "|$WILDCARD_INDEX|").split(".", "|")
+        val path = lookupPathSegments(dictionaryLookupPath)
         val values = StringProviders.getFor(pattern, this, path)
 
         return values.map(::StringValue).firstNotNullOfOrNull { value ->
@@ -469,6 +469,41 @@ data class Resolver(
     }
 
     private fun lastLookupKey(): String? = dictionaryLookupPath.substringAfterLast(".").takeIf(String::isNotBlank)
+
+    internal fun lookupPathSegments(path: String): List<String> {
+        val result = ArrayList<String>(lookupPathsSeenSoFar.size)
+
+        var i = 0
+        var start = 0
+        while (i < path.length) {
+            when {
+                path[i] == '.' -> {
+                    flushSegment(path, start, i, result)
+                    start = ++i
+                }
+
+                path.isWildcardAt(i) -> {
+                    flushSegment(path, start, i, result)
+                    result.add(WILDCARD_INDEX)
+                    i += 3
+                    start = i
+                }
+
+                else -> i++
+            }
+        }
+
+        if (start < path.length) result.add(path.substring(start))
+        return result
+    }
+
+    private fun String.isWildcardAt(i: Int): Boolean {
+        return i + 2 < length && this[i] == '[' && this[i + 1] == '*' && this[i + 2] == ']'
+    }
+
+    private fun flushSegment(path: String, start: Int, end: Int, out: MutableList<String>): Unit {
+        if (end > start) out.add(path.substring(start, end))
+    }
 }
 
 private fun ExactValuePattern.hasPatternToken(): Boolean {
--- a/core/src/main/kotlin/io/specmatic/core/Resolver.kt
+++ b/core/src/main/kotlin/io/specmatic/core/Resolver.kt
@@ -54,6 +54,7 @@ data class Resolver(
     val lookupPathsSeenSoFar: Set<String> = setOf(),
     val cycleMarker: String = "",
     val maxTestRequestCombinations: Int = Int.MAX_VALUE,
+    val maxRandomArraySize: Int = 3,
 ) {
     constructor(facts: Map<String, Value> = emptyMap(), mockMode: Boolean = false, newPatterns: Map<String, Pattern> = emptyMap()) : this(CheckFacts(facts), mockMode, newPatterns)
     constructor() : this(emptyMap(), false)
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/io/specmatic/core/utilities/EarlyResult.kt
```diff
@@ -27,3 +27,15 @@ inline fun <Item, Value, Failure> Iterable<Item>.firstSuccessOrFailures(evaluate
...
     return EarlyResult.Failures(failures)
 }
+
+inline fun <Item> Iterable<Item>.firstSuccessOrFailures(evaluate: (Item) -> io.specmatic.core.Result): EarlyResult<Item, io.specmatic.core.Result.Failure> {
+    val failures = mutableListOf<io.specmatic.core.Result.Failure>()
+
+    for (item in this) {
+        val result = evaluate(item)
+        if (result !is io.specmatic.core.Result.Failure) return EarlyResult.FirstSuccess(item)
+        failures.add(result)
```
sahilm:
```suggestion
        if (result.isSuccess()) {
            return EarlyResult.FirstSuccess(item)
        }
        when (result) {
            is io.specmatic.core.Result.Failure -> failures.add(result)
            else -> error("unexpected type of Result") // please adapt this accordingly
        }
```
I suggest this change to make the code malleable and less error prone. Ideally it would be good to have an exhaustive pattern match over all types of Results and handle each one.

core/src/main/kotlin/io/specmatic/core/utilities/EarlyResult.kt
```diff
@@ -27,3 +27,15 @@ inline fun <Item, Value, Failure> Iterable<Item>.firstSuccessOrFailures(evaluate
...
+    }
+
+    return EarlyResult.Failures(failures)
```
sahilm, line 41:
```suggestion
    return EarlyResult.Failures(failures.toList())
```
I suggest limiting the scope of `mutableList` because it's a source of bugs to pass around mutable things. If it helps it's good to avoid creating a `mutableList` in the first place.

core/src/main/kotlin/io/specmatic/core/Resolver.kt
```diff
@@ -469,6 +471,41 @@ data class Resolver(
     }
 
     private fun lastLookupKey(): String? = dictionaryLookupPath.substringAfterLast(".").takeIf(String::isNotBlank)
+
+    internal fun lookupPathSegments(path: String): List<String> {
```
sahilm, line 475: What's the the performance gain achieved by this? It's replacing much simpler code.

core/src/test/kotlin/io/specmatic/core/ResolverKtTest.kt
```diff
@@ -42,5 +50,56 @@ internal class ResolverKtTest {
                 Arguments.of("Schema", "[*]", "Schema[*]"),
             )
         }
+
+        @JvmStatic
+        fun lookupPathSegmentsProvider(): Stream<Arguments> {
```
sahilm, line 55: good test 👍

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityPerfTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class TestBackwardCompatibilityPerfTest {
+    @Test
+    fun `backward compatibility check performance should not regress`() {
+        val specFile = File(javaClass.getResource("/openapi/multi_res_perf.yaml")!!.toURI())
+        val oldFeature = OpenApiSpecification.fromFile(specFile.canonicalPath).toFeature()
+        val newFeature = OpenApiSpecification.fromFile(specFile.canonicalPath).toFeature()
+
+        var results: Results? = null
+        assertTimeout(ofSeconds(1)) {
```
sahilm:
```suggestion
        assertTimeoutPreemptively(5.seconds.toJavaDuration()) {
```
1. We want to abort execution preemptively  so that the test suite doesn't slow down.
2. Bumped up to 5 seconds because CI can be very slow at times. On my machine tests are finishing in 225 ms.

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityPerfTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import io.specmatic.conversions.OpenApiSpecification
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertTimeout
+import org.junit.jupiter.api.Test
+import java.io.File
+import java.time.Duration.ofSeconds
+import kotlin.system.measureTimeMillis
+
+class TestBackwardCompatibilityPerfTest {
```
sahilm, line 11: This is very good test and a very good start to a perf test suite. There's a slight risk that this will be flaky because of slow CI environment but I think it's ok to wait and watch. Longer term we would want to run perf tests after basic correctness tests.

sahilm (approved):
Great work! There's a few suggested changes that I would like you to consider and reason about. Let's have a second pair of eyes on this by @joelrosario too before merging.

core/src/main/kotlin/io/specmatic/core/utilities/EarlyResult.kt
StarKhan6368: Sounds good, addressed with an exhaustive `when` check

core/src/main/kotlin/io/specmatic/core/utilities/EarlyResult.kt
StarKhan6368: Yes, that is basically what is occurring here through implicit upcasting. It is created as a MutableList internally for convenience but is passed around as a List

core/src/main/kotlin/io/specmatic/core/Resolver.kt
StarKhan6368: Fairly significant because this is a hot path, With the same specs, runtime dropped from **1m36s to 1m20s**, which is about a **16–17% improvement** overall, The impact should scale with generation volume, and especially with nesting depth of the schema.

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityPerfTest.kt
StarKhan6368: Makes sense, I have updated the test accordingly

PR #2494, opened 2026-05-12
chore: update reporter version to 0.3.19, it basically fixes the status of wip tests by marking them as other and displays them using an hourglass icon in the html report

yogeshnikam671:


jaydeepk (approved):
LGTM

PR #2495, opened 2026-05-12
Implement CTRF report generation for backward compatibility

StarKhan6368:
**What**: Implement CTRF report generation for backward compatibility

**Why**:
- CTRF will let users and orgs consume compatibility results in a richer format instead of relying on console
- Reporting alignment with test and mock workflows will provide a more consistent user experience.

**Note**: 
The generation of the report is currently feature flagged under `SPECMATIC_BCC_REPORT` until the implementation is complete

**Checklist**:

- [x] Unit Tests
- [x] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Demo video (share link)

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
+++ b/core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
@@ -7,17 +7,23 @@ import io.specmatic.core.utilities.capitalizeFirstChar
 import io.specmatic.core.value.NullValue
 import io.specmatic.core.value.Value
 import io.specmatic.reporter.ctrf.model.CtrfBackwardCompatibilityRecord
+import io.specmatic.reporter.ctrf.model.CtrfReport
 
 private const val BCC_REPORT_DIR_SUFFIX = "backward_compatibility"
 private const val SPECMATIC_BCC_REPORT_FLAG = "SPECMATIC_BCC_REPORT"
 
 fun testBackwardCompatibility(older: Feature, newer: Feature): Results {
+    return testBackwardCompatibilityWithReport(older, newer).first
+}
+
+internal fun testBackwardCompatibilityWithReport(older: Feature, newer: Feature): Pair<Results, CtrfReport?> {
     val startTime = System.currentTimeMillis()
     val records = OpenApiBackwardCompatibilityChecker(older, newer).run()
     val endTime = System.currentTimeMillis()
 
-    generateBackwardCompatibilityReport(records, startTime, endTime)
-    return records.toBackwardCompatibilityResults()
+    val result = records.toBackwardCompatibilityResults()
+    val report = generateBackwardCompatibilityReport(records, startTime, endTime)
+    return Pair(result, report)
 }
 
 fun List<CtrfBackwardCompatibilityRecord>.toBackwardCompatibilityResults(): Results {
@@ -28,11 +34,11 @@ fun List<CtrfBackwardCompatibilityRecord>.toBackwardCompatibilityResults(): Resu
         }
 }
 
-private fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long) {
-    if (!Flags.getBooleanValue(SPECMATIC_BCC_REPORT_FLAG)) return
+private fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long): CtrfReport? {
+    if (!Flags.getBooleanValue(SPECMATIC_BCC_REPORT_FLAG)) return null
     val reportOperations = BccReportGenerator().generateReportOperations(records)
     val reportDir = loadSpecmaticConfigOrDefault(getConfigFileName()).getReportDirPath(BCC_REPORT_DIR_SUFFIX).toFile()
-    ReportGenerator.generateReportBcc(
+    return ReportGenerator.generateReportBcc(
         endTime = endTime,
         startTime = startTime,
         reportDir = reportDir,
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+class BccReportGenerator {
+    fun generateReportOperations(tests: List<CtrfBackwardCompatibilityRecord>): List<BaseBccCoverageReportOperation> {
+        return tests.operationSources().groupByOperation().map { (groupKey, sources) ->
+            reportOperationFrom(groupKey, sources)
+        }
+    }
+
+    private fun reportOperationFrom(groupKey: BccOperationGroupKey, sources: List<BccOperationSource>): BaseBccCoverageReportOperation {
+        val tests = sources.map { it.test }
+        return BccCoverageReportOperation(
```
sahilm: `BccCoverageReportOperation` should we have the word coverage in here?

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
```diff
@@ -1,12 +1,47 @@
...
+    return filterIsInstance<OpenApiBackwardCompatibilityCheckRecord>()
+        .groupBy { it.operations }.values
+        .fold(Results()) { acc, recordsForOperation ->
+            acc.plus(Results(recordsForOperation.map { it.compatResult }).distinct())
+        }
+}
+
+private fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long) {
+    if (!Flags.getBooleanValue(SPECMATIC_BCC_REPORT_FLAG)) return
+    runCatching {
```
sahilm: Why are we `runCatching`? Can we just let it fail instead?

core/src/test/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckerTest.kt
```diff
@@ -0,0 +1,67 @@
+package io.specmatic.core
+
+import io.specmatic.conversions.OpenApiSpecification
+import io.specmatic.toViolationReportString
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class OpenApiBackwardCompatibilityCheckerTest {
+    @Test
+    fun `run should remove failure reasons from generated records without hiding mismatch`() {
```
sahilm, line 10: Why do we need to remove the failure reason? Mostly just a question for my understanding.

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityKtTest.kt
```diff
@@ -4501,6 +4567,134 @@ paths:
...
+            assertThat(summary.path("failed").asInt()).isEqualTo(1)
+            assertThat(reportJson.path("results").path("tests").size()).isEqualTo(5)
+            assertThat(reportJson.path("results").path("extra").path("reportType").asText()).isEqualTo("BackwardCompatibility")
+            assertThat(reportJson.path("results").path("extra").path("specmaticConfigPath").asText()).isEqualTo(configFile.canonicalPath)
+
+            assertThat(operations.size()).isEqualTo(2)
+            assertThat(operations.toString()).contains("\"method\":\"GET\"")
+            assertThat(operations.toString()).contains("\"method\":\"POST\"")
+            assertThat(operations.toString()).contains("\"backwardCompatibilityResult\":\"Compatible\"")
+            assertThat(operations.toString()).contains("\"backwardCompatibilityResult\":\"Incompatible\"")
```
sahilm: Please expand this to assert on the full body of the json so that we have one test that covers what entire json report should look like. You can use a fixture json that you expect, load it up into the data classes and then compare the data classes if you like.

sahilm (approved):
Looks good. Left some comments for you.

core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
StarKhan6368: Good point, i've renamed the class to not include the word `coverage`

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
StarKhan6368: There’s no specific reason,
I've removed `runCatching`, since the operation is already controlled by a feature flag

core/src/test/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckerTest.kt
StarKhan6368: FailureReasons are what determine whether a result is considered fluffy or not, and if it is, what the fluff level is. There appears to be a significant amount of functionality that has the side effect of eliminating fluffy errors, such as the .distinct family of methods in the Results class. This ensures that what we send is never removed or filtered, as we know that whether fluffy or not, these are the exact failures we want to show.

core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityKtTest.kt
StarKhan6368: Performing an exact equality assertion is challenging in this case because many fields are dynamically generated and may differ between executions, such as the start, end, and UUIDs throughout the entire ctrfReport. We can do it if it feels necessary, I've asserted the important sections instead

joelrosario (approved):

PR #2496, opened 2026-05-12
Annotate BCC breadcrumbs with source spec:line:col

sahilm:
## Summary

1. Annotate every breaking-change breadcrumb in BCC output with a `(spec:line:col)` source location, so downstream tooling can render linter-style diagnostics (eventually SARIF) that point a user at the exact line that broke.
2. Results are structured with `SourceLocation` which can later be used by SARIF exporter.
3. When the failure is associated with:
    1. Additions (example new fields): the new yaml source location is provided.
    1. Type changes: the new yaml source location is provided.
    1. Removals (fields, complete operations): the old yaml source location is provided.
    This means that tooling will need to have access to both old and new specs.   

## TODO Items
1. Handle cases where refs point to external files.
4. Handle spec files that are in JSON.
5. Test out and figure out what's the behaviour with YAML anchors and YAML merging/composition.


sahilm:
I'm deliberately avoiding addressing the merge conflict before PR review because there's already a bit of complexity. The merge conflict is due to a single isolated change that I had to pull to make the tests deterministic (generate only 1 entry per ListPattern). Also there's a potential bug fix for that in here.

Once the review is done and specifics around this PR are addressed, addressing the merge conflict will be simple.

core/src/main/kotlin/io/specmatic/core/FlagsBased.kt
```diff
@@ -13,7 +13,8 @@ data class FlagsBased(
     val negativePrefix: String,
     val allPatternsAreMandatory: Boolean,
     val useFuzzyMatching: Boolean,
-    val maxTestRequestCombinations: Int
+    val maxTestRequestCombinations: Int,
+    val maxRandomArraySize: Int = 3,
```
joelrosario: why was this added?

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -723,20 +727,30 @@ class OpenApiSpecification(
...
+                    val queryParameterPointers = buildParameterPointers(openApiPath, httpMethod, "query", pathItem.parameters, openApiOperation.parameters, methodContext)
+                    val requestHeaderPointers = buildParameterPointers(openApiPath, httpMethod, "header", pathItem.parameters, openApiOperation.parameters, methodContext)
+                    val pathParameterPointers = buildParameterPointers(openApiPath, httpMethod, "path", pathItem.parameters, openApiOperation.parameters, methodContext)
                     val specmaticPathParam = toSpecmaticPathParam(
                         openApiPath = openApiPath,
                         parameters = parameters,
                         otherPathPatterns = allPathPatternsGroupedByMethod[httpMethod].orEmpty(),
-                        collectorContext = methodContext
+                        collectorContext = methodContext,
+                        parameterPointers = pathParameterPointers.mapValues { "${it.value}/name" }
```
joelrosario, line 779: path parameters can be reffed out, so name may not exist

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -825,7 +841,8 @@ class OpenApiSpecification(
                             specification = specificationPath,
                             protocol = protocol,
                             specType = SpecType.OPENAPI,
-                            operationMetadata = operationMetadata
+                            operationMetadata = operationMetadata,
+                            sourceLocations = sourceLocationsFromMap()
```
joelrosario: can't this be called once instead of each time a scenario is called, as it's not doing anything new

joelrosario (changes requested):
- Paths, requests and responses, and others can be reffed out. Does the sourceLocation logic take this into account?
- In the case of allOf, the schemas within the allOf are internally reconstituted into a single schema. Does the sourceLocation logic take this into account?

core/src/main/kotlin/io/specmatic/core/FlagsBased.kt
sahilm: This is removed now but I've maintained the bug fix in [7ec0bf6](https://github.com/specmatic/specmatic/pull/2496/commits/7ec0bf635d199071fbbc1a820aa7469f4b5709ac). The bug is that in main right now the max array size is off-by-one. 2 instead of 3.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
sahilm: Thanks. I've fixed this and validated that it works in the integration tests. Details in the top level comment: https://github.com/specmatic/specmatic/pull/2496#issuecomment-4571860585.

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
sahilm: Yes indeed. Fixed.

sahilm:
> * Paths, requests and responses, and others can be reffed out. Does the sourceLocation logic take this into account?
> * In the case of allOf, the schemas within the allOf are internally reconstituted into a single schema. Does the sourceLocation logic take this into account?

I've validated all of these cases in https://github.com/specmatic/specmatic/blob/a02f70dd7977ebb77e7efacd4631bd4033b7a59e/core/src/test/kotlin/io/specmatic/core/TestBackwardCompatibilityKtTest.kt#L4987. I've gone through both `orders_old.yaml` and `order_new.yaml` and made sure that reported sourceLocations are correct for all reffed out entities, allOf, anyOf and oneOf. Thanks to your review I found an issue with reffed out path params that I fixed in https://github.com/specmatic/specmatic/pull/2496/changes/77e110d2c47cf8f820c0ecad965ae04db5dfc480. Please see `pathScopePointer`.

joelrosario (approved):

PR #2499, opened 2026-05-13
chore: remove match field from CtrfTestMetadata

vedubhat:
Reason to change: 
 
Removing the addition of 'match' field from CtrfTestMetadata so that it does not appear in ctrf-report as it is not used in html-reporter and insights


yogeshnikam671 (approved):
LGTM

PR #2500, opened 2026-05-13
fix: show WIP count under WIP in the logs logged by Junit execution listener

yogeshnikam671:


yogeshnikam671:
Approved by @jaydeepk 
He could not put an approval himself because of some issue

PR #2505, opened 2026-05-18
Mark an operation as changed in the BCC report if it was edited in the new spec

sahilm:
Mark an operation as changed in the BCC report if it was edited in the new spec by fingerprinting scenarios and comparing old vs new scenario.

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
@@ -97,7 +97,8 @@ class OpenApiSpecification(
     private val dictionary: Dictionary = loadDictionary(openApiFilePath, specmaticConfig.getDictionary(), strictMode),
     private val logger: LogStrategy = io.specmatic.core.log.logger,
     private val parseCollectorContext: CollectorContext = CollectorContext(),
-    private val exampleDirPaths: List<String> = emptyList()
+    private val exampleDirPaths: List<String> = emptyList(),
+    private val changeTrackingSource: OpenApiChangeTrackingSource? = null,
 ) : IncludedSpecification, ApiSpecification {
     private val extensibleQueryParams: Boolean = specmaticConfig.getExtensibleQueryParams()
     private val preferEscapedSoapAction: Boolean = specmaticConfig.getEscapeSoapAction()
@@ -335,7 +336,40 @@ class OpenApiSpecification(
                 lenientMode = lenientMode,
                 logger = logger,
                 parseCollectorContext = collectorContext,
-                exampleDirPaths = exampleDirPaths
+                exampleDirPaths = exampleDirPaths,
+                changeTrackingSource = OpenApiChangeTrackingSource(
+                    yamlContent = preprocessedYaml,
+                    openApiFilePath = openApiFilePath.replace("\\", "/"),
+                    lenientMode = lenientMode,
+                ),
+            )
+        }
+
+        internal fun fromYAMLForChangeTracking(
+            yamlContent: String,
+            openApiFilePath: String,
+            specmaticConfig: SpecmaticConfig = SpecmaticConfig(),
+            strictMode: Boolean = false,
+            lenientMode: Boolean = false,
+            exampleDirPaths: List<String> = emptyList(),
+        ): OpenApiSpecification {
+            val parseResult: SwaggerParseResult =
+                OpenAPIV3Parser().readContents(
+                    yamlContent,
+                    null,
+                    resolveFullyReferences(),
+                    openApiFilePath.replace("\\", "/")
+                )
+            val parsedOpenApi: OpenAPI = parseResult.openAPI
+                ?: throw ContractException("Could not parse contract $openApiFilePath for change tracking, please validate the syntax using https://editor.swagger.io")
+
+            return OpenApiSpecification(
+                openApiFilePath,
+                parsedOpenApi,
+                specmaticConfig = specmaticConfig,
+                strictMode = strictMode,
+                lenientMode = lenientMode,
+                exampleDirPaths = exampleDirPaths,
             )
         }
 
@@ -367,6 +401,12 @@ class OpenApiSpecification(
                 it.isResolve = true
                 it.isResolveRequestBody = true
                 it.isResolveResponses = true
+            }
+        }
+
+        private fun resolveFullyReferences(): ParseOptions {
+            return resolveExternalReferences().also {
+                it.isResolveFully = true
             }
         }
 
@@ -526,7 +566,8 @@ class OpenApiSpecification(
             specmaticConfig = specmaticConfig,
             strictMode = strictMode,
             protocol = protocol,
-            exampleDirPaths = exampleDirPaths
+            exampleDirPaths = exampleDirPaths,
+            changeTrackingSource = changeTrackingSource,
         )
 
         return Pair(feature, rootContext.toCollector().toResult())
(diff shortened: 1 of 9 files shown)


sahilm:
1. The failing tests are deliberate and are handled either by https://github.com/specmatic/specmatic/pull/2506 or https://github.com/specmatic/specmatic/pull/2507. Need a discussion to decide on the approach.
2. Due to a bug we may mark a BCC failure as UNCHANGED which would lead to a poor UX. Should we do a set union of all failures and changed scenarios and mark all of them as changed? Or alternatively, only compute fingerprints for passing scenarios while short circuiting evaluation of failing scenarios marking them as changed immediately?

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityChecker.kt
```diff
@@ -3,40 +3,48 @@ package io.specmatic.core
...
-                val responseResults = validateResponseCompatibility(requestFamily, requestResults).also(::addAll)
+                val changeStatus = operationChangeStatus(requestFamily)
+                val requestResults = validateRequestCompatibility(requestFamily, changeStatus).also(::addAll)
+                val responseResults = validateResponseCompatibility(requestFamily, requestResults, changeStatus).also(::addAll)
                 logOperationResult(requestResults.plus(responseResults))
             }
         }
     }
 
+    private fun operationChangeStatus(requestFamily: RequestFamily): ChangeStatus {
```
StarKhan6368: Not sure if i'm following correctly, shouldn't operations changeStatus include all 5 identifiers, path, method, reqCT, status and resCT ?

core/src/main/kotlin/io/specmatic/core/ScenarioFingerprint.kt
```diff
@@ -0,0 +1,30 @@
...
+
+data class ScenarioFingerprint(
+    val status: Int,
+    val requestContentType: String?,
+    val responseContentType: String?,
+    val httpRequestPattern: HttpRequestPattern,
+    val httpResponsePattern: HttpResponsePattern,
+) {
+    companion object {
+        fun from(scenario: Scenario): ScenarioFingerprint = ScenarioFingerprint(
```
StarKhan6368, line 21: Not really an issue, but status, requestContentType and responseContentType are already part of httpRequestPattern and httpResponsePattern

StarKhan6368 (changes requested):

StarKhan6368:
I would prefer PR [#2507](https://github.com/specmatic/specmatic/pull/2507).

Reason: OpenAPI semantics live in the resolved schema graph, not in the raw $ref shape. For BCC, we want to know whether the operation meaning has changed, and the change-tracking parse using fullyResolve appears to be cycle-safe for self-recursion, optional recursion, and mutual recursion based on the cases I've tested manually

PR [#2506](https://github.com/specmatic/specmatic/pull/2506) is feasible, but it spreads reference-awareness across multiple pattern classes. If a new pattern type or nested reference path is missed later, and we would need to own recursive traversal and cycle handling across those pattern types ourselves ?

But I do not see explicit tests covering these recursive cases in [#2507](https://github.com/specmatic/specmatic/pull/2507), Could we please add tests for self-recursion, optional recursion, and mutual recursion and others, when feasible? We cannot always trust the parse, it has regressed before.

And for the `UNCHANGED` case for failure, I'd prefer short-circuit failing scenarios as changed, although I still need to understand what conditions or what bug is causing this failure

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityChecker.kt
sahilm: Thanks for spotting this. I've fixed it and added comprehensive tests both unit and one integration style tests that covers the entire surface area.

core/src/main/kotlin/io/specmatic/core/ScenarioFingerprint.kt
sahilm: Want to leave this as-is for clarity 👍

sahilm:
> I would prefer PR [#2507](https://github.com/specmatic/specmatic/pull/2507).
> 
> Reason: OpenAPI semantics live in the resolved schema graph, not in the raw $ref shape. For BCC, we want to know whether the operation meaning has changed, and the change-tracking parse using fullyResolve appears to be cycle-safe for self-recursion, optional recursion, and mutual recursion based on the cases I've tested manually
> 
> PR [#2506](https://github.com/specmatic/specmatic/pull/2506) is feasible, but it spreads reference-awareness across multiple pattern classes. If a new pattern type or nested reference path is missed later, and we would need to own recursive traversal and cycle handling across those pattern types ourselves ?
> 
> But I do not see explicit tests covering these recursive cases in [#2507](https://github.com/specmatic/specmatic/pull/2507), Could we please add tests for self-recursion, optional recursion, and mutual recursion and others, when feasible? We cannot always trust the parse, it has regressed before.
> 
I've merged #2507 into this PR and added tests for recursive scenarios.

> And for the `UNCHANGED` case for failure, I'd prefer short-circuit failing scenarios as changed, although I still need to understand what conditions or what bug is causing this failure

There's no such bug at the moment. We can ignore this. This was just for discussion.

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityChecker.kt
```diff
@@ -3,40 +3,55 @@ package io.specmatic.core
...
 import io.specmatic.core.pattern.ContractException
 import io.specmatic.core.pattern.IgnoreUnexpectedKeys
+import io.specmatic.reporter.internal.dto.bcc.ChangeStatus
 import io.specmatic.test.asserts.toFailure
 import java.util.Collections
 import java.util.IdentityHashMap
 
 class OpenApiBackwardCompatibilityChecker(private val oldFeature: Feature, private val newFeature: Feature) {
     private val newScenariosByMethodAndReqContentType = newFeature.scenarios.groupBy { it.method to it.requestContentType }
+    private val oldChangeTrackingScenariosByPathAndMethod = oldFeature.scenariosForChangeTracking()
```
StarKhan6368, line 13: How will this function when the path changes between the old and new specifications? For example, the meaning might remain the same, but /order/{id} could change to /order/{orderId} while still using the same data type

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -335,7 +336,40 @@ class OpenApiSpecification(
...
+        internal fun fromYAMLForChangeTracking(
+            yamlContent: String,
+            openApiFilePath: String,
+            specmaticConfig: SpecmaticConfig = SpecmaticConfig(),
+            strictMode: Boolean = false,
+            lenientMode: Boolean = false,
+            exampleDirPaths: List<String> = emptyList(),
+        ): OpenApiSpecification {
+            val parseResult: SwaggerParseResult =
+                OpenAPIV3Parser().readContents(
```
StarKhan6368, line 357: Can we add a test for fromYAMLForChangeTracking itself, to confirm parsed YAML is resolved fully? I found coverage for similar ref cases, but not a direct assertion on this path, as it's external code

core/src/main/kotlin/io/specmatic/core/Feature.kt
```diff
@@ -185,8 +185,19 @@ data class Feature(
     val strictMode: Boolean = false,
     val exampleStore: ExampleStore = ExampleStore.empty(),
     val protocol: SpecmaticProtocol,
-    val exampleDirPaths: List<String> = emptyList()
+    val exampleDirPaths: List<String> = emptyList(),
+    val changeTrackingSource: OpenApiChangeTrackingSource? = null,
 ): IFeature {
+    private val cachedScenariosForChangeTracking: List<Scenario> by lazy {
```
StarKhan6368, line 191: Should we revert to using scenarios? It seems to give the caller a false sense of security, if it's not present, it's simply not available

StarKhan6368 (changes requested):

StarKhan6368 (approved):
Looks good to me

PR #2509, opened 2026-05-19
Fix BCC change tracking for external refs

sahilm:
## Summary
- Materialize the newer OpenAPI feature's change-tracking scenarios before the BCC command checks out the base branch.
- Add a command-level regression test that runs through the real temporary git repository, branch switch, base checkout, BCC report generation, and embedded HTML report JSON assertion.

## Bug
During the CLI backward compatibility flow, the newer spec is parsed while the worktree is on the current branch. The command then stashes local changes and checks out the base branch before loading the older spec and running the comparison.

OpenAPI change tracking resolves external refs lazily when scenariosForChangeTracking() is first evaluated. If that first evaluation happens after the checkout, the newer fingerprints can resolve external schema files from the base branch instead of the current branch. That makes external-schema-only changes look unchanged in the BCC report, even though the current branch changed the referenced schema.

## Fix
The command now calls scenariosForChangeTracking() on the newer Feature immediately after parsing it and before any branch checkout. This materializes the newer change-tracking scenarios while the current branch's external files are still present in the worktree, so later comparison/report generation uses the correct fingerprints.

## Verification
- env JAVA_HOME=/opt/homebrew/Cellar/openjdk@17/17.0.18/libexec/openjdk.jdk/Contents/Home ./gradlew :specmatic-executable:test --tests 'application.BackwardCompatibilityCheckCommandV2Test'

### DIFF
--- a/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
+++ b/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
@@ -251,6 +251,12 @@ abstract class BackwardCompatibilityCheckBaseCommand(
                     }
 
                     val newer = getFeatureFromSpecPath(specFilePath)
+                    // The newer feature is parsed while the worktree has the current branch files, but below we
+                    // checkout the base branch before running the comparison. OpenAPI change tracking resolves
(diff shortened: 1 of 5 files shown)


StarKhan6368 (approved):
Looks good to me

PR #2510, opened 2026-05-19
Add changed qualifier to BCC CTRF operations

sahilm:
We add `Changed` as a qualifier so that it shows up as a pill under remarks in the HTML report. I've tested the report produced by the integration test and it shows up correctly. The `Changed`/`Unchanged` status is left untouched right now in the remaining CTRF so this is a non breaking change.

## Companion PRs

- specmatic/specmatic-reporter#136 — adds the `CHANGED` value to `CtrfOperationQualifiers`.
- specmatic/specmatic-html-reporter#106 — capitalizes qualifier pills in the HTML report so `changed` renders as `Changed`.

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
+++ b/core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
@@ -19,28 +19,20 @@ class BccReportGenerator {
 
     private fun reportOperationFrom(groupKey: BccOperationGroupKey, sources: List<BccOperationSource>): BaseBccReportOperation {
         val tests = sources.map { it.test }
-        val changeStatus = operationChangeStatus(tests)
         return BccReportOperation(
             tests = tests,
             operation = groupKey.operation,
             specConfig = groupKey.specConfig,
-            qualifiers = operationQualifiersFrom(tests, changeStatus),
+            qualifiers = operationQualifiersFrom(tests),
             compatibility = CtrfOperationCompatibility(
-                changeStatus = changeStatus,
+                changeStatus = operationChangeStatus(tests),
                 result = backwardCompatibilityResultFrom(tests),
             ),
         )
     }
 
-    private fun operationQualifiersFrom(
-        tests: List<CtrfBackwardCompatibilityRecord>,
-        changeStatus: ChangeStatus,
-    ): List<CtrfOperationQualifiers> {
-        val testQualifiers = tests.flatMap { it.operationQualifiers }
-        val changeQualifier = listOfNotNull(
-            CtrfOperationQualifiers.CHANGED.takeIf { changeStatus == ChangeStatus.CHANGED }
-        )
-        return (testQualifiers + changeQualifier).distinct()
+    private fun operationQualifiersFrom(tests: List<CtrfBackwardCompatibilityRecord>): List<CtrfOperationQualifiers> {
+        return tests.flatMap { it.operationQualifiers }.distinct()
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
```diff
@@ -19,20 +19,28 @@ class BccReportGenerator {
...
             ),
         )
     }
 
-    private fun operationQualifiersFrom(tests: List<CtrfBackwardCompatibilityRecord>): List<CtrfOperationQualifiers> {
-        return tests.flatMap { it.operationQualifiers }.distinct()
+    private fun operationQualifiersFrom(
+        tests: List<CtrfBackwardCompatibilityRecord>,
+        changeStatus: ChangeStatus,
+    ): List<CtrfOperationQualifiers> {
```
StarKhan6368: - Can we add it to the operationQualifiers of OpenApiBackwardCompatibilityCheckRecord instead?
- This can remain a simple flatMap -> distinct

core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
```diff
@@ -19,20 +19,28 @@ class BccReportGenerator {
...
+        val changeStatus = operationChangeStatus(tests)
         return BccReportOperation(
             tests = tests,
             operation = groupKey.operation,
             specConfig = groupKey.specConfig,
-            qualifiers = operationQualifiersFrom(tests),
+            qualifiers = operationQualifiersFrom(tests, changeStatus),
             compatibility = CtrfOperationCompatibility(
-                changeStatus = operationChangeStatus(tests),
+                changeStatus = changeStatus,
```
StarKhan6368: I assume this PR does not address the changes discussed regarding the removal of the compatibility object. For now, we are duplicating the changeStatus function here and in the qualifiers ?

core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
```diff
@@ -19,20 +19,28 @@ class BccReportGenerator {
...
+        changeStatus: ChangeStatus,
+    ): List<CtrfOperationQualifiers> {
+        val testQualifiers = tests.flatMap { it.operationQualifiers }
+        val changeQualifier = listOfNotNull(
+            CtrfOperationQualifiers.CHANGED.takeIf { changeStatus == ChangeStatus.CHANGED }
+        )
+        return (testQualifiers + changeQualifier).distinct()
     }
 
     private fun operationChangeStatus(tests: List<CtrfBackwardCompatibilityRecord>): ChangeStatus {
```
StarKhan6368, line 38: This will then no longer be required, and we will not need to include changeStatus in the CtrfBackwardCompatibilityRecord interface

core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
sahilm: Yes this is correct. In this PR we will duplicate. Will follow up with the changes in the structure later.

core/src/main/kotlin/io/specmatic/core/report/BccReportGenerator.kt
sahilm: Done.

StarKhan6368 (approved):
Looks good to me

PR #2515, opened 2026-05-25
Remove feature toggle. Make bcc report feature live.

sahilm:


StarKhan6368:
If we proceed with the release, it will not work properly with the backward compatibility check in its current state right ? Because report generation occurs per specification, and the last specification checked overrides all previously generated reports

sahilm:
> If we proceed with the release, it will not work properly with the backward compatibility check in its current state right ? Because report generation occurs per specification, and the last specification checked overrides all previously generated reports

Ack. Closing this PR. Will fix and then reopen.

PR #2516, opened 2026-05-25
Support WIP in backward compatibility checks and reports

sahilm:
Makes WIP behave in backward compatibility checks (BCC) exactly as it does in the test/mock CTRF/HTML reports.

## Problem
WIP scenarios (`@WIP` tag / `[WIP]` row / OpenAPI `WIP` operation tag → `Scenario.ignoreFailure`) were **filtered out of BCC entirely** — `OpenApiBackwardCompatibilityChecker` skipped them for both execution and change tracking, so a WIP scenario produced no record at all. The existing "a breaking WIP scenario should not break BCC" behaviour worked only because the scenario was skipped. And even when records were constructed, the CTRF status ignored WIP, so a WIP incompatibility counted as `failed`.

## Change
- **Execute & change-track WIP** (`OpenApiBackwardCompatibilityChecker`): removed the `ignoreFailure` filters from execution and from both change-tracking maps. WIP scenarios now run and are change-tracked like any other operation.
- **Keep WIP out of the verdict** (`TestBackwardCompatibility.toBackwardCompatibilityStatuses`): WIP records are excluded from the pass/fail fold, so a breaking WIP scenario produces a real failure record without breaking the verdict.
- **Report like test/mock** (`OpenApiBackwardCompatibilityCheckRecord`): `isWip = scenario.ignoreFailure` plus a `wip` tag (no name prefix — WIP is conveyed via tag metadata). Combined with the reporter companion, WIP BCC tests report as `other` (retaining the real raw status), render with the hourglass icon, and are counted under `other`/Wip rather than `failed`/Errors.
- Adds a WIP operation to the `bcc_integration` fixtures and asserts the behaviour end to end, plus a `WipMatrix` test suite over the command's console output covering WIP/non-WIP × compatible/breaking, including the WIP tag being added or removed in the new spec (the new spec's tag governs the exemption).

## Companion PRs
This is the main PR. It depends on:
- specmatic/specmatic-reporter#140 — WIP-aware CTRF status + `extra.wip` metadata
- specmatic/specmatic-html-reporter#119 — regenerated example fixture + e2e

## Validation
- `:specmatic-core` BCC/record tests, the reporter test, and `:specmatic-executable` BCC tests pass.
- Browser-validated against the html reporter: the WIP `/promotions` operation shows Incompatible + `wip`/`changed` pills, the summary surfaces a `Wip: 2` bucket (no Errors), and the drill-down renders both executions as `Wip` (the failing one retains its incompatibility message, the passing one has none).

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
+++ b/core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
@@ -22,11 +22,14 @@ data class OpenApiBackwardCompatibilityCheckRecord(
     override val branch: String? = scenario.sourceRepositoryBranch
     override val specification: String = scenario.specification ?: feature.path
 
+    override val isWip: Boolean = scenario.ignoreFailure
+
     // TODO: Need actual positive variation from generatedScenario
-    override val name: String = scenario.fullApiDescription
+    override val name: String = if (isWip) "WIP: ${scenario.fullApiDescription}" else scenario.fullApiDescription
     override val message: String = compatResult.reportString()
     override val operations: Set<APIOperation> = toOpenApiOperation(scenario)
     override val tags: List<String> = buildList {
+        if (isWip) add("wip")
         add("status:${scenario.status}")
         add("method:${scenario.method.lowercase()}")
         add("path:${convertPathParameterStyle(scenario.path)}")
@@ -40,7 +43,7 @@ data class OpenApiBackwardCompatibilityCheckRecord(
     }
 
     override val operationQualifiers: List<CtrfOperationQualifiers> = buildList {
-        if (scenario.ignoreFailure) add(CtrfOperationQualifiers.WIP)
+        if (isWip) add(CtrfOperationQualifiers.WIP)
         if (changeStatus == ChangeStatus.CHANGED) add(CtrfOperationQualifiers.CHANGED)
     }
 
--- a/core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
+++ b/core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
@@ -25,7 +25,7 @@ data class OpenApiBackwardCompatibilityCheckRecord(
     override val isWip: Boolean = scenario.ignoreFailure
 
     // TODO: Need actual positive variation from generatedScenario
-    override val name: String = if (isWip) "WIP: ${scenario.fullApiDescription}" else scenario.fullApiDescription
+    override val name: String = scenario.fullApiDescription
     override val message: String = compatResult.reportString()
     override val operations: Set<APIOperation> = toOpenApiOperation(scenario)
     override val tags: List<String> = buildList {
--- a/core/src/main/kotlin/io/specmatic/core/Results.kt
+++ b/core/src/main/kotlin/io/specmatic/core/Results.kt
@@ -6,7 +6,14 @@ data class Results(val results: List<Result> = emptyList()) {
     fun hasResults(): Boolean = results.isNotEmpty()
 
     fun hasFailures(): Boolean = results.any { it is Result.Failure }
-    fun success(): Boolean = if(hasResults()) successCount > 0 && failureCount == 0 else true
+    fun success(): Boolean =
+        if (hasResults()) results.none { it is Result.Failure && !it.shouldBeIgnored() } else true
+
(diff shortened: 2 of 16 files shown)


core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
```diff
@@ -22,11 +22,14 @@ data class OpenApiBackwardCompatibilityCheckRecord(
     override val branch: String? = scenario.sourceRepositoryBranch
     override val specification: String = scenario.specification ?: feature.path
 
+    override val isWip: Boolean = scenario.ignoreFailure
+
     // TODO: Need actual positive variation from generatedScenario
-    override val name: String = scenario.fullApiDescription
+    override val name: String = if (isWip) "WIP: ${scenario.fullApiDescription}" else scenario.fullApiDescription
```
StarKhan6368: Is there a specific reason we are prefixing them with WIP? I don't see us doing this for test and mock records.

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
```diff
@@ -28,6 +28,7 @@ internal fun testBackwardCompatibilityWithReport(older: Feature, newer: Feature)
 
 fun List<CtrfBackwardCompatibilityRecord>.toBackwardCompatibilityStatuses(): Results {
     return filterIsInstance<OpenApiBackwardCompatibilityCheckRecord>()
+        .filter { !it.scenario.ignoreFailure }
```
StarKhan6368: Just to clarify, we have agreed that WIP failures will not be displayed on the console ? as they're being filtered out here

StarKhan6368:
Have we considered how WIP will behave when there are changes between the old and new specifications? In my mind, I was anticipating the behavior to be somewhat as follows, but i don't see this happening in code
| Old spec state | New spec state | Behavior | Reasoning |
|---|---|---|---|
| WIP | WIP | Ignore compatibility failures | Both versions are unstable, so there is no stable contract to protect. |
| WIP | Stable | Ignore compatibility failures | This is a promotion from WIP to stable. The new version becomes the first stable contract. |
| Stable | WIP | Do not ignore compatibility failures, Report them as an error. | A stable API should not be able to bypass compatibility checks by being marked WIP later. |
| Stable | Stable | Run normal compatibility checks | Both versions are part of the stable contract. |

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
sahilm: As discussed keeping this so that this shows up as WIP in console output too.

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
sahilm: Great point. Changing this so that WIP scenarios show up in console output but count as failures.

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityCheckRecord.kt
sahilm: Changed my mind. Removed the WIP prefix because it doesn't show up in console output anyway.

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
sahilm: This is done.

core/src/main/kotlin/io/specmatic/core/Results.kt
```diff
@@ -6,7 +6,14 @@ data class Results(val results: List<Result> = emptyList()) {
     fun hasResults(): Boolean = results.isNotEmpty()
 
     fun hasFailures(): Boolean = results.any { it is Result.Failure }
-    fun success(): Boolean = if(hasResults()) successCount > 0 && failureCount == 0 else true
+    fun success(): Boolean =
```
StarKhan6368: Should we maintain the current Results semantics and introduce a BCC-only helper. This way, we make an additive change without disrupting existing callers that depend on the current Results contract

In its current state, this change contradicts the output from other helpers

sahilm: Done. I've fixed this so that there's a bcc specific helper.

StarKhan6368 (approved):
Looks good to me, thanks!

PR #2518, opened 2026-05-27
Fix BCC CTRF report dropping all but the last checked spec

sahilm:
## Problem

- When the backward-compatibility check (`backward-compatibility-check`) checks more than one changed spec with `SPECMATIC_BCC_REPORT=true`, the generated CTRF/HTML report only ever contains the **last** spec checked. Every earlier spec is silently dropped from `results.summary.extra.executionDetails`, `results.tests`, and the summary counts.

- Also found and fixed another issue related to hook dropping passing scenarios. See test `ctrf report includes compatible specs when hook validates mixed failures` for details.

### Root cause

The report was regenerated and written **once per spec**, inside the per-spec loop: `checkBackwardCompatibility` → `testBackwardCompatibility` → `testBackwardCompatibilityWithReport` → `generateBackwardCompatibilityReport` → `ReportGenerator.generateReportBcc`, which builds the report from only the current spec's records and **overwrites** the same `index.html` / CTRF file each iteration. Last write wins.

### Companion PRs
- https://github.com/specmatic/enterprise/pull/306

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
+++ b/core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
@@ -40,6 +40,7 @@ fun List<CtrfBackwardCompatibilityRecord>.toBackwardCompatibilityStatuses(): Res
 }
 
 fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long): CtrfReport? {
+    if (records.isEmpty()) return null
     if (!Flags.getBooleanValue(SPECMATIC_BCC_REPORT_FLAG)) return null
     val reportOperations = BccReportGenerator().generateReportOperations(records)
     val reportDir = loadSpecmaticConfigOrDefault(getConfigFileName()).getReportDirPath(BCC_REPORT_DIR_SUFFIX).toFile()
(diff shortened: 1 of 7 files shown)


sahilm:
Companion enterprise PR: specmatic/enterprise#306 (adapts the gRPC/async/GraphQL BCC commands to the new `checkBackwardCompatibility` return type).

gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=io.specmatic
-version=2.46.2-SNAPSHOT
+version=2.46.3-SNAPSHOT
```
StarKhan6368: Is there a reason for bumping the version?

core/src/main/kotlin/io/specmatic/core/TestBackwardCompatibility.kt
```diff
@@ -34,7 +39,8 @@ fun List<CtrfBackwardCompatibilityRecord>.toBackwardCompatibilityStatuses(): Res
         }
 }
 
-private fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long): CtrfReport? {
+fun generateBackwardCompatibilityReport(records: List<CtrfBackwardCompatibilityRecord>, startTime: Long, endTime: Long): CtrfReport? {
+    if (records.isEmpty()) return null
```
StarKhan6368, line 43: It would have skipped the generation regardless, but an explicit check is fine

StarKhan6368 (approved):
Looks good to me, thanks!

PR #2521, opened 2026-05-28
Distinguish and name positive request variations in the backward-compatibility CTRF report

sahilm:
1. Request variations get fully described in the bcc report. These will later be used to explain the work specmatic does while running bcc. Right now it brings the report to parity with the mock usage report's scenario names but this is not going to be the final form.
2. Removed `newBasedOnBacwardsCompatability`.


core/src/main/kotlin/io/specmatic/core/Scenario.kt
```diff
@@ -503,8 +504,9 @@ data class Scenario(
 
                 val newResponsePattern: HttpResponsePattern = this.httpResponsePattern.withResponseExampleValue(rowValue, resolver)
 
+                val requestStatus = if (flagsBased.requestSchemaOnly) 200 else httpResponsePattern.status
```
StarKhan6368, line 507: Would be cleaner to pass a flag into the function with a derived default.
Instead of manipulating the status by sending an incorrect value,
we could modify the `newBasedOn` function to accept a flag that determines whether to read or create.

core/src/main/kotlin/io/specmatic/core/OpenApiBackwardCompatibilityChecker.kt
```diff
@@ -55,7 +55,7 @@ class OpenApiBackwardCompatibilityChecker(private val oldFeature: Feature, priva
 
     private fun generatePositiveVariations(scenarios: List<Scenario>): List<Scenario> {
         return scenarios.flatMap { scenario ->
-            scenario.withoutExamples().generateBackwardCompatibilityScenarios()
+            scenario.withoutExamples().generateBackwardCompatibilityScenarios().map { it.value }
```
StarKhan6368, line 58: Need handling for HasFailure and HasException, as invoking $.value on these instances will result in an exception being thrown

core/src/main/kotlin/io/specmatic/core/pattern/AnyOfPattern.kt
```diff
@@ -147,6 +147,11 @@ data class AnyOfPattern(
     ): Sequence<ReturnValue<Pattern>> {
         val newPatterns = delegate.newBasedOn(row, resolver)
 
+        // Backward-compatibility checks compare each anyOf branch's request schema on its own; the
+        // synthesized union-of-branches object is not a distinct branch and only produces a redundant
+        // (and indistinguishable) positive variation, so skip it on that path.
+        if (resolver.prioritisedRequestCombinationsOnly) return newPatterns
```
StarKhan6368, line 153: Worth being careful with the early return here: overlap-only anyOf compatibility gaps exist today, and while they’re not caused only by this flag, skipping the combined variant here makes that behavior easier to miss.

core/src/main/kotlin/io/specmatic/core/FlagsBased.kt
```diff
@@ -15,6 +15,14 @@ data class FlagsBased(
...
     val maxTestRequestCombinations: Int,
     val randomArraySize: Int? = null,
+    val prioritisedRequestCombinationsOnly: Boolean = false,
+    // When set, request generation always follows the valid path regardless of the scenario's
+    // response status. Normally a 4xx response (400/422) drives invalid-request generation
+    // (omitting required query/header params) to exercise the error path. Backward-compatibility
+    // checks only compare request schemas, so they must always get a valid request -- otherwise two
+    // identical specs whose representative response is a 4xx would generate param-omitting requests
+    // and be falsely flagged as request-incompatible.
+    val requestSchemaOnly: Boolean = false,
```
StarKhan6368, line 25: We can remove this if we agree with the comment made for newBasedOn
This doesn't feel like it belongs in FlagBased

core/src/main/kotlin/io/specmatic/core/pattern/CombinationSpec.kt
```diff
@@ -84,6 +85,9 @@ class CombinationSpec<ValueType>(
             if(prioritisedGenerations.size == maxCombinations)
                 return@sequence
 
+            if(prioritisedOnly)
```
StarKhan6368, line 88: Stopping after the prioritized prefix does not seem sound for backward-compatibility checks. 
A skipped non-prioritized combination can still be a real old-valid request that the new spec could reject

Example, this old schema accepts all 4 combinations of `paymentMethod x currency`:
```yaml
type: object
required: [paymentMethod, currency]
properties:
  paymentMethod:
    type: string
    enum: [card, upi]
  currency:
    type: string
    enum: [USD, INR]
```

If it changes to:
```yaml
oneOf:
  - type: object
    required: [paymentMethod, currency]
    additionalProperties: false
    properties:
      paymentMethod:
        type: string
        enum: [card]
      currency:
        type: string
        enum: [USD]
  - type: object
    required: [paymentMethod, currency]
    additionalProperties: false
    properties:
      paymentMethod:
        type: string
        enum: [upi]
      currency:
        type: string
        enum: [INR]
```

> Also I'm not sure the name conveys the meaning. To me, prioritization means reordering, but this skips some elements

StarKhan6368 (changes requested):

StarKhan6368:
We are not removing the old `newBasedOn` code paths in this PR.
@joelrosario need you help with the review of combinatorial optimizations related to `anyOf` and `CombinationSpec` specifically

sahilm:
Closing this PR as this feature is quite complex to get right v/s the value we get out of it.

PR #2523, opened 2026-05-29
Fix change-status flipping siblings on a path-parameter change

sahilm:
The otherPathPatterns were breaking change tracking because they were unrelated. Just removed them while fingerprinting.


core/src/main/kotlin/io/specmatic/core/ScenarioFingerprint.kt
```diff
@@ -20,7 +20,12 @@ data class ScenarioFingerprint(
             status = scenario.status,
             requestContentType = scenario.requestContentType,
             responseContentType = scenario.responseContentType,
-            httpRequestPattern = scenario.httpRequestPattern,
+            // Drop the ambient sibling-path context so that an operation's change status reflects
+            // only its own contract - otherwise a path-pattern change in one operation would flip
+            // the change status of every other operation sharing the same HTTP method.
+            httpRequestPattern = scenario.httpRequestPattern.let { request ->
```
StarKhan6368, line 26: It would be preferable to avoid manual copying and allow the scenario to isolate self before a fingerprint is created.
1. We are aiming to reduce the use of copy and instead use explicit functions.
2. We will soon implement a similar specificity matching for content-type in headers

Neither of these points is a blocker at the moment tho, they can be addressed later when content-type specificity is implemented

StarKhan6368 (approved):
Looks good to me

PR #2533, opened 2026-06-01
Annotate BCC breadcrumbs with source locations across external $refs

sahilm:
## Summary

Follow-up to #2496. In this PR we add support for external $refs. Most of the diff is new tests that cover various cases involved. I've also tested this against https://github.com/specmatic/labs-contracts where the BCC report now accurately boils down all breakages to a single change in a common spec. We can use this make the report succinct and boil down errors while still showing the user all the tested cases.

Please note this PR doesn't contain the `->` markers leading from spec to external ref. That will follow in another PR.


StarKhan6368 (approved):
Looks good to me, @joelrosario could you please take a look as well, since you reviewed the base implementation last time

joelrosario (approved):

PR #2550, opened 2026-06-09
Normalize filesystem spec paths for federated CTRF reports

jaydeepk:
## Summary
- implements #2559 
- normalize filesystem-backed spec paths to Git-repo-relative paths for CTRF stitching
- apply that normalization in OpenAPI and WSDL loading paths and in CTRF spec config generation
- preserve `repository = null` for filesystem CTRF metadata instead of emitting an empty string
- add focused coverage and mock-report assertions for OpenAPI spec paths

## Problem
In a federated repo setup, provider specs live inside the provider repository instead of a separate central contracts repository.

For Insights stitching to work, the service-side CTRF report and the provider repo central-contract-style report must converge on the same canonical spec identity.
For filesystem-backed specs, that means:
- the reported `specification` path must be Git-repo-relative
- filesystem providers must not carry misleading empty-string repository metadata

Before this branch:
- filesystem spec paths could remain config-relative or loader-relative
- provider CTRF metadata could emit `repository = ""`
- that made federated provider identity stitching brittle across Specmatic, Reporter, and Insights

## Fix
This branch introduces `normalizeFilesystemSpecificationPath(...)` and threads it through the relevant filesystem-backed flows.

Applied in:
- `parseContractFileToFeature(...)`
- `OpenApiSpecification`
- WSDL feature creation
- `SpecmaticConfigV1V2Common.getCtrfSpecConfig(...)`
- `SpecmaticConfigV3Impl.getCtrfSpecConfig(...)`

Behavior:
- when the source is `filesystem` and the resolved spec file is inside a Git repo, the reported spec path becomes repo-root-relative
- non-filesystem providers keep their existing spec paths unchanged
- filesystem CTRF metadata now preserves `repository = null` instead of coercing it to `""`

## Scope note
This normalization currently applies to the OpenAPI and WSDL / SOAP flows, plus CTRF spec-config generation for filesystem-backed specs.

It does **not** add equivalent normalization for Gherkin specs in this branch.
The Gherkin path remains unchanged and is out of scope for this PR.

## Validation
- added `CtrfSpecificationPathTest` covering:
  - OpenAPI filesystem normalization
  - WSDL filesystem normalization
  - V1/V2 and V3 CTRF config normalization
  - non-filesystem sources remaining unchanged
- updated regression assertions so filesystem CTRF repository metadata is `null`
- added focused OpenAPI coverage and mock-usage assertions for expected spec paths


StarKhan6368:
I believe we can narrow this down.

For the configuration-backed flows of interest, `SpecmaticJUnitSupport.kt`, `StubCommand.kt`, and the programmatic stub path in `api.kt` all utilize `loadContracts()` for filesystem sources. 

This indicates that any modifications to `LocalFileSystemSource.loadContracts()` should address the repo-root-relative path scenario without needing to alter the parser or reporting code.

The broader normalization in the feature/report layer seems to only affect direct single-file execution with no config. If we do not want that behavior, I’d recommend the surgical change in `LocalFileSystemSource` instead, what do you think @jaydeepk ?

jaydeepk:
Thats a great suggestion @StarKhan6368 
As discussed, I have removed nomalization from Feature, OpenApiSpecification and added it in LocalFileSystemSource.
We would however still need to retain  nomalization in the V3 and V1V2 confg classes

core/src/main/kotlin/io/specmatic/stub/report/OpenApiMockUsage.kt
```diff
@@ -39,8 +35,4 @@ class OpenApiMockUsage(
             absoluteCoverage = coverageReportOperations.calculateAbsoluteCoverage(),
         )
     }
-
-    fun ctrfSpecConfigs(): List<CtrfSpecConfig> {
```
StarKhan6368, line 43: Can we remove `ctrfSpecConfigsFrom`? It looks like it's no longer in use across the repositories

core/src/main/kotlin/io/specmatic/core/CtrfSpecificationPath.kt
```diff
@@ -0,0 +1,29 @@
+package io.specmatic.core
+
+import io.specmatic.core.git.SystemGit
+import java.io.File
+
+internal fun normalizeFilesystemSpecificationPath(
+    specificationPath: String,
+    sourceProvider: String?,
```
StarKhan6368: Minor nitpick, but now that all callers can provide the SourceProvider enum, could we switch to using that instead of a free-form string?

core/src/main/kotlin/io/specmatic/core/config/v3/SpecmaticConfigV3Impl.kt
```diff
@@ -236,13 +237,21 @@ data class SpecmaticConfigV3Impl(val file: File? = null, val specmaticConfig: Sp
...
+            }
 
         return CtrfSpecConfig(
             protocol = protocol,
             specType = specType,
-            specification = specPathFromConfig.orEmpty(),
-            sourceProvider = source?.toProviderType()?.name ?: SourceProvider.filesystem.name,
-            repository = source?.getGit()?.url.orEmpty(),
+            specification = normalizedSpecificationPath.orEmpty(),
+            sourceProvider = source?.toProviderType()?.name.orEmpty(),
```
StarKhan6368: To prevent unexpected issues, would it be okay to reintroduce the default filesystem fallback here or default to null instead ?

core/src/main/kotlin/io/specmatic/stub/report/OpenApiMockUsage.kt
jaydeepk: Yes will remove this

core/src/main/kotlin/io/specmatic/core/config/v3/SpecmaticConfigV3Impl.kt
jaydeepk: Agreed ..bringing it back

core/src/main/kotlin/io/specmatic/core/CtrfSpecificationPath.kt
jaydeepk: Yes ..makes sense

StarKhan6368 (approved):
Looks good to me, thanks!

PR #2551, opened 2026-06-09
Remove SPECMATIC_BCC_REPORT feature flag

sahilm:
Source locations are now always added and the BCC report is always
generated, removing the now-unconditional feature toggle.


StarKhan6368 (approved):
Looks good to me, thanks!

PR #2567, opened 2026-06-12
Remove problematic usages of java.net.URL

sahilm:
1. Java's `java.net.URL` is fundamentally broken: its `equals()`/`hashCode()` perform **blocking DNS resolution**, making URL comparison slow, unreliable, and network-dependent. JDK 20 deprecated its `URL(String)` constructors for this reason.
2. Usages of `java.net.URL` have been replaced with `java.net.URI`. `io.ktor.http.Url` was evaluated but there's no need to use it in this codebase.
3. Added test coverage where it was missing so that we ensure a behavior preserving change.
4. Removed the test only usage of `OkHttp`. We can use Ktor's Http client when needed. It's not needed here.


core/src/main/kotlin/io/specmatic/core/utilities/ResolvedWebSource.kt
```diff
@@ -65,16 +65,7 @@ data class ResolvedWebSource(
...
-                val url = URL(specPath)
-                if (url.protocol.isNotBlank()) {
-                    throw ContractException("Web source specifications must be relative paths, but got \"$specPath\"")
-                }
-            } catch (_: MalformedURLException) {
-                // Relative paths are not valid URLs and are allowed here.
-            }
-
-            val parsed = runCatching { URI(specPath) }.getOrNull()
+            val parsed = runCatching { URI(specPath.replace('\\', '/')) }.getOrNull()
```
sahilm, line 67: `.replace('\\', '/'))` is needed to cover the windows path case. For other cases use of URL isn't needed.

core/src/main/kotlin/io/specmatic/core/SpecmaticConfigV1V2Common.kt
```diff
@@ -1659,12 +1658,9 @@ data class Source(
...
 
-            cachedWebSpec ?: try {
-                val url = URL(specPath)
-                sourceBaseDir.resolve("web").resolve(url.host).resolve(url.path.removePrefix("/")).canonicalFile
-            } catch (_: MalformedURLException) {
-                sourceBaseDir.resolve(specPath).canonicalFile
-            }
+            cachedWebSpec ?: specPath.toHttpUrlOrNull()?.let { url ->
+                sourceBaseDir.resolve("web").resolve(url.host).resolve(url.encodedPath.removePrefix("/")).canonicalFile
+            } ?: sourceBaseDir.resolve(specPath).canonicalFile
```
sahilm: This was uncovered by tests. Added coverage in this pr.

core/src/main/kotlin/io/specmatic/conversions/Postman.kt
```diff
@@ -213,11 +212,12 @@ fun postmanItemRequest(request: JSONObjectValue): Pair<String, HttpRequest> {
...
     }
 
-    val httpRequest = HttpRequest(method, url.path, headers, body, query, formFields, formData)
+    val path = uri.rawPath
+    val httpRequest = HttpRequest(method, path, headers, body, query, formFields, formData)
     return Pair(baseURL, httpRequest)
 }
 
-internal fun urlFromPostmanValue(urlValue: Value): URL {
+internal fun urlFromPostmanValue(urlValue: Value): URI {
```
sahilm, line 220: Opted to make this change so that we can remove the import of `java.net.URL`.

core/src/main/kotlin/io/specmatic/conversions/Postman.kt
```diff
@@ -169,10 +168,10 @@ fun postmanItemResponse(responseItem: JSONObjectValue): HttpResponse {
 
 fun postmanItemRequest(request: JSONObjectValue): Pair<String, HttpRequest> {
     val method = request.getString("method")
-    val url = urlFromPostmanValue(request.jsonObject.getValue("url"))
+    val uri = urlFromPostmanValue(request.jsonObject.getValue("url"))
 
-    val baseURL = "${url.protocol}://${url.authority}"
-    val query: Map<String, String> = url.query?.split("&")?.map { it.split("=").let { parts -> Pair(parts[0], parts[1]) } }?.fold(emptyMap()) { acc, entry -> acc.plus(entry) }
+    val baseURL = "${uri.scheme}://${uri.rawAuthority}"
+    val query: Map<String, String> = uri.rawQuery?.split("&")?.map { it.split("=").let { parts -> Pair(parts[0], parts[1]) } }?.fold(emptyMap()) { acc, entry -> acc.plus(entry) }
```
ketan, line 174: Looks good overall. The only thing I'd double-check is validating `raw[query|authority]` against `query|authority`, since that's the only area I can see potentially causing problems

core/src/main/kotlin/io/specmatic/core/URLParts.kt
```diff
@@ -34,4 +36,12 @@ class URLParts(url: String) {
...
-}
\ No newline at end of file
+}
+
+// URL schemes are case-insensitive (RFC 3986), so compare them ignoring case.
+fun String?.isHttpScheme(): Boolean = equals("http", ignoreCase = true) || equals("https", ignoreCase = true)
+
+// URI.host is null for authorities Java rejects as hostnames (e.g. "backend_service"), even though the
+// authority is retained and the URL is reachable. Fall back to the raw authority, minus any userinfo/port.
+fun URI.hostOrAuthority(): String =
```
sahilm, line 46: I've added this because URL allows hostnames like `backend_service` even though the standard prohibits underscores in hostnames. Need some help to decide if we want this or not.

joelrosario (approved):

PR #2584, opened 2026-06-17
feat: validate changed externalised examples and validate all externalised examples of a changed spec as part of backward compatibility check

### LINKED ISSUE #2571: Externalised examples support in BCC
Acceptance Criteria:

- BCC should validate the changed externalised examples
- BCC should validate all the externalised examples whose spec/s has changed.

yogeshnikam671:
Addresses - https://github.com/specmatic/specmatic/issues/2571


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -110,74 +115,66 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
         return generateSequence(path, Path::getParent).find {
-            it.pathString.endsWith("_examples") || it.pathString.endsWith("_tests")
+            it.pathString.endsWith("_examples")
         }
     }
 
-    override fun areExamplesValid(feature: IFeature, which: String): Boolean {
+    override fun File.isExternalisedExample(): Boolean = isFile && extension == "json" && runCatching { ScenarioStub.readFromFile(this) }.isSuccess
+
+    override fun getExternalExampleValidationResults(feature: IFeature): ValidationResults {
```
StarKhan6368: Why are we re-evaluating the directories ? The method `getExternalExampleDirectories` is already called, why aren't we passing the directories down ? What happens if they become out of sync for any reason ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -508,3 +823,34 @@ abstract class BackwardCompatibilityCheckBaseCommand(
         }
     }
 }
+
+private fun <T> captureCompatibilityOutput(block: () -> T): Pair<T, String> {
```
StarKhan6368: What is the intent of capturing the logger output instead of displaying it in real-time ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -508,3 +823,34 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+            }
+
+            capturedOutput.appendLine(renderedMessage)
+        }
+    }
+
+    return ThreadSafeLog(NonVerbose(CompositePrinter(listOf(printer))))
+}
+
+private fun String.shouldBeSkippedFromCapturedCompatibilityOutput(): Boolean {
```
StarKhan6368: There must be a more effective way to disable inline-example warnings, have you thought about adjusting the specmatic config so that `getIgnoreInlineExampleWarnings` returns true?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -454,26 +647,148 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+        logger.log(summaryTitle.prependIndent(TWO_INDENTS))
+        if (!hasValidationErrors && exampleValidationResults.isNotEmpty()) {
+            logger.log("Validated ${exampleValidationResults.size} externalised example(s).".prependIndent(TWO_INDENTS))
+        }
+        logger.log(Results(exampleValidationResults.values.toList()).summary().prependIndent(TWO_INDENTS))
+        logger.log("=".repeat(summaryTitle.length).prependIndent(TWO_INDENTS))
+    }
+
+    private fun logUnloadableExamples(unusedExamples: Set<String>) {
+        logger.boundary()
```
StarKhan6368, line 795: Could the indentation be better managed by using `logger.withIndentation`? This way, it would be included in the visualization but excluded from the actual log, and you don't have to track how many idents to use next

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -454,26 +647,148 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+    }
+
+    private fun ProcessedSpec.hasBackwardCompatibilityFailure(): Boolean {
+        if (backwardCompatibilityResult.successExcludingIgnorableFailures()) return false
+        return failedVerdictMessage(this, hook, effectiveStrictMode, effectiveBaseBranch).first == CompatibilityResult.FAILED
+    }
+
+    private fun ProcessedSpec.hasExternalExampleValidationFailure(): Boolean {
+        if (!backwardCompatibilityResult.successExcludingIgnorableFailures()) return false
+        return !externalisedExamples.validationResults.success || externalisedExamples.unloadableExamples.isNotEmpty()
```
StarKhan6368, line 704: Does this imply that specifications which have failed both BCC and example validation are not considered as `specExternalExampleValidationFailureCount` ?

yogeshnikam671: The behaviour is:
- If spec itself is incompatible, do not validate the examples. (No point in validating examples against an incompatible spec. But open to discuss, not locked in yet.)
- If spec is compatible then validate the examples so confirm those are still valid.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: Addressed, mostly.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -110,74 +117,59 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+
+            ExternalisedExampleBackwardCompatibilityEvaluation(
+                validationResults = exampleValidationModule.validateExamples(feature, examples = externalExampleFiles),
+                directories = externalExampleDirectories
+                    .map { it.canonicalFile.relativeTo(repoDir).invariantSeparatorsPath }
+                    .toSet(),
+                files = externalExampleFiles
+                    .map { it.canonicalFile.relativeTo(repoDir).invariantSeparatorsPath }
+                    .toSet(),
+                unloadableExamples = feature.loadExternalisedExamplesAndListUnloadableExamples().second
```
StarKhan6368, line 168: By default this will only look at test example folders not all of them, this will cause drift between this and files, as files will be a superset, Consider using a single source for both

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -110,74 +117,59 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+            File(it).isExternalisedExample()
+        }.mapNotNull {
+            getParentExamplesDirectory(Paths.get(it))
+        }.distinct().flatMap { examplesDir ->
+            val specBasePath = examplesDir.parent.resolve(
+                examplesDir.fileName.toString().removeSuffix("_examples")
+            )
 
-        return result
+            findSpecFiles(specBasePath)
```
StarKhan6368, line 132: Possible bug in `findSpecFiles`, because its concatenating won't the files search space become `[nameyaml, namejson, nameyml, etc]` instead of `[name.yaml, name.json, name.yml, etc]`, possible tests missing for other extensions as well then

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -82,28 +83,57 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+            logger.log("$newLine No specs were changed, skipping the check.$newLine")
             logger.log(CompatibilityReport.emptyReport())
             return 0
         }
 
+        val changedExternalisedExampleFiles = filesChangedInCurrentBranch.filter {
+            File(it).isExternalisedExample()
+        }.toSet()
+        val changedSpecFiles = changedSpecFiles(filesChangedInCurrentBranch, changedExternalisedExampleFiles)
+        val filesReferringToChangedSpecFiles = getSpecsReferringTo(filesChangedInCurrentBranch).filter {
```
StarKhan6368, line 114: Shouldn't `getSpecsReferringTo` take `changedSpecFiles` instead of `filesChangedInCurrentBranch` ? possibly false match of refs if the example name matches

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
StarKhan6368: Got it, Why is example validation executed before backward compatibility checks then ? If this is intentional can we re-order that to avoid both computation and confusion of why we're discarding its result

StarKhan6368 (changes requested):
One thing to note, unloadable external examples can be skipped because changed files are validated before being classified as externalised examples. Example-only changes may pass without reporting invalid examples.

The refactor is reasonable, but separating it into its own PR would have made review and regression detection easier

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
StarKhan6368: Seems like most of it has remained, but its fine if it can't be done, not a blocker to me

yogeshnikam671:
Closing the PR. Will be addressed as part of a separate PR.

PR #2585, opened 2026-06-18
feat: add spec-level coverage metrics to CTRF reports

### LINKED ISSUE #2557: Calculate spec-level coverage and mock usage
Add logic to calculate coverage and mock usage percentages per spec instead of only at build level. Reuse existing percentage logic, but run it on each grouped spec’s operations. Add executionDetails.metrics to the CTRF JSON output. Keep existing top-level build metrics unchanged. Add tests for single-spec, multi-spec, mock report, and zero-operation edge cases. 

              
Acceptance Criteria -

- [x] CTRF report includes spec-level coverageMetrics inside each executionDetails entry
- [ ] Spec-level apiCoverage ,absoluteCoverage, totalOperations, coveredOperations & totalOperationsWithFilters are calculated by grouping operations per spec and reusing the existing coverage percentage logic
- [x] Existing top-level build metrics remain unchanged
- [ ] Spec matching works correctly across relative & absolute paths that refer to the same spec
- [ ] Specs with no matching operations do not break CTRF generation
- [x] Contract and mock/stub report flows continue generating CTRF successfully
- [ ] Unit tests cover single-spec, multi-spec, mock-flow, and zero-operation scenarios

Sushant092003:
**What**: Add spec-level coverage metrics to CTRF execution details so each spec in a build can carry its own apiCoverage and absoluteCoverage values, while keeping the existing build-level coverage values unchanged.

**Why**: Insights needs spec-scoped coverage data from the CTRF report instead of relying only on build-level percentages or recomputing coverage later. Since the build history view is scoped by spec, using only overall build coverage would be misleading. The CTRF output therefore needs to include coverage calculated per spec.

**How**: Implemented spec-level grouping of coverage operations in specmatic-core, calculated apiCoverage and absoluteCoverage for each grouped spec using the existing coverage calculation logic, and wrapped each spec’s CtrfSpecConfig, operations, and coverage metrics into CoverageReportSpecification. Updated CTRF report generation to pass these wrapper objects through to specmatic-reporter, and added unit and integration coverage around multi-spec execution details and final CTRF JSON output.

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/report/CoverageReportOperationExtensions.kt
+++ b/core/src/main/kotlin/io/specmatic/core/report/CoverageReportOperationExtensions.kt
@@ -5,28 +5,40 @@ import io.specmatic.reporter.internal.dto.coverage.CoverageStatus
 import io.specmatic.reporter.internal.dto.coverage.OmittedStatus
 import kotlin.math.roundToInt
 
+fun List<BaseCoverageReportOperation>.coveredOperationsCount(): Int {
+    return this.count { operation ->
+        operation.eligibleForCoverage && operation.coverageStatus == CoverageStatus.COVERED
+    }
+}
+
+fun List<BaseCoverageReportOperation>.totalOperationForCoverageIncludingFilters(): List<BaseCoverageReportOperation> {
+    return this.filter { it.eligibleForCoverage }
+}
+
+fun List<BaseCoverageReportOperation>.totalOperationsForCoverage(): List<BaseCoverageReportOperation> {
+    return this.filter { operation ->
+        operation.eligibleForCoverage || operation.omittedStatus == OmittedStatus.EXCLUDED
+    }
+}
+
 fun List<BaseCoverageReportOperation>.calculateCoverage(): Int {
-    val coverageReportOperations = this.filter { it.eligibleForCoverage }
+    val coverageReportOperations = this.totalOperationForCoverageIncludingFilters()
     if (coverageReportOperations.isEmpty()) {
         return 0
     }
 
-    val coveredOperationCount = coverageReportOperations.count { it.coverageStatus == CoverageStatus.COVERED }
+    val coveredOperationCount = coverageReportOperations.coveredOperationsCount()
     return ((coveredOperationCount.toDouble() / coverageReportOperations.size) * 100).roundToInt()
 }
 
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/io/specmatic/core/report/CoverageReportOperationExtensions.kt
```diff
@@ -5,28 +5,40 @@ import io.specmatic.reporter.internal.dto.coverage.CoverageStatus
...
+    val denominatorOperations = this.totalOperationsForCoverage()
 
     if (denominatorOperations.isEmpty()) {
         return 0
     }
 
-    val coveredOperationCount = denominatorOperations.count { operation ->
-        operation.eligibleForCoverage && operation.coverageStatus == CoverageStatus.COVERED
-    }
+    val coveredOperationCount = denominatorOperations.coveredOperationsCount()
```
yogeshnikam671, line 42: @jaydeepk  is better equipped to review this.

core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
```diff
@@ -14,18 +14,19 @@ import java.io.File
...
         startTime: Long,
         endTime: Long,
-        specConfigs: List<CtrfSpecConfig>,
         coverage: Int? = null,
         actuatorEnabled: Boolean? = null,
         absoluteCoverage: Int? = null,
         reportDir: File,
         toolName: String = "Specmatic ${VersionInfo.describe()}",
     ) {
+        val specConfigs = coverageReportSpecifications.map { it.specConfig }
```
yogeshnikam671: why does this variable exist?

core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
```diff
@@ -14,18 +14,19 @@ import java.io.File
...
         if(isCtrfSpecConfigsValid(specConfigs).not()) return
 
+        val coverageReportOperations = coverageReportSpecifications.flatMap { it.coverageReportOperations }
```
yogeshnikam671: same question, why does this variable exist?

yogeshnikam671 (changes requested):

yogeshnikam671 (approved):
LGTM

PR #2589, opened 2026-06-19
validate snapshots before we do any heavy lifting

ketan:
ensure that we validate snapshot deps and fail the build in case of snapshot reps early on in the build


yogeshnikam671 (approved):
LGTM

PR #2590, opened 2026-06-19
External examples schema

SaachiKaup:
JSON Schema for external Examples.

Currently tracks the work done by Sufiyan and Joel in `external_examples.yaml`.

### Notes to discuss

For backward compatibility reasons, `path` field in `http-request` is not mandatory in the code. It is however _strongly_ recommended that it be present. For both humans and LLMs. If they are the intended audience, we make it mandatory. 

Thus, in current implementation, the field is mandatory. Both normally, and when defined inside a partial.


yogeshnikam671 (approved):
Fixture schema looks good to me.

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+          type: string
+        wait:
+          nullable: true
+          type: string
+        timeout:
+          nullable: true
+          type: string
+        retry:
+          nullable: true
+          $ref: "#/components/schemas/Retry"
```
StarKhan6368, line 187: The reference appears to be invalid, I do not see a scheme named Retry

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+        type:
+          type: string
+        wait:
+          nullable: true
+          type: string
+        timeout:
+          nullable: true
+          type: string
+        retry:
+          nullable: true
```
StarKhan6368: Should we update the `openapi` version to `3.1.x` to ensure that these nullables are recognized as valid ?

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
       required:
         - name
         - filename
+
+    HttpFixture:
+      type: object
+      properties:
+        type:
+          type: string
```
StarKhan6368, line 179: Shouldn't the type be const `http` ? Ideally, we would have a `oneOf` with a discriminator setup, which we can still do, For now, there will only be one entry, and it will be when the discriminator key is `http`

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+        wait:
+          nullable: true
+          type: string
```
StarKhan6368: This is a ISO 8601 duration format, have we checked if there's a regex we can utilize here if its simple, ideally we would have `duration` format, but we don't support that yet

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
         - name
         - filename
+
+    HttpFixture:
+      type: object
+      properties:
+        type:
+          type: string
+        wait:
+          nullable: true
```
StarKhan6368: Why is `wait` marked nullable ? i don't see this reflecting in-code

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+          type: string
+        timeout:
+          nullable: true
```
StarKhan6368: Why is `timeout` marked nullable ? i don't see this reflecting in-code

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+          type: string
```
StarKhan6368: Same here, ISO 8601 duration format

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+        retry:
+          nullable: true
+          $ref: "#/components/schemas/Retry"
+        executeFor:
+          nullable: true
```
StarKhan6368: Why is `executeFor` marked nullable ? i don't see this reflecting in-code

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+          $ref: "#/components/schemas/FixtureExecuteFor"
+        http-request:
+          $ref: "#/components/schemas/RequestSchema"
+        http-response:
```
StarKhan6368, line 192: Shouldn't `http-request` and `http-response` be marked as required keys ? as they don't have defaults unlike the rest of the fields

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+          nullable: true
+          type: string
+        retry:
+          nullable: true
+          $ref: "#/components/schemas/Retry"
+        executeFor:
+          nullable: true
+          $ref: "#/components/schemas/FixtureExecuteFor"
+        http-request:
+          $ref: "#/components/schemas/RequestSchema"
```
StarKhan6368: This can't be the same schema as that of the example, if you glance over executor code, it has exactly fields `baseUrl`, `path`, `method`, `headers` abd `body`

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
...
+        http-response:
+          $ref: "#/components/schemas/ResponseSchema"
```
StarKhan6368: This can't be the same schema as that of the example, if you glance over executor code, it has exactly fields `status` and `body` only

core/src/main/resources/schemas/external_example.yaml
```diff
@@ -137,3 +147,25 @@ components:
       required:
         - name
         - filename
+
+    HttpFixture:
```
StarKhan6368, line 175: Seems unused, shouldn't this be used in `before` and `after` array

core/src/main/resources/schemas/external_examples.schema.json
```diff
@@ -0,0 +1,297 @@
+{
```
StarKhan6368, line 1: Seems like this already has proper fixtures schemas, please update the other file as well, with similar content

core/src/main/resources/schemas/external_examples.schema.json
```diff
@@ -0,0 +1,297 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
StarKhan6368, line 2: I don't see any default values defined for the fields here. If I recall correctly, the goal is to publish this so users can validate and modify the examples within their IDEs and LLMs can create examples. Therefore, we should include default values and ideally descriptions for the fields as well

StarKhan6368 (changes requested):

StarKhan6368:
@SaachiKaup Could we implement a test that ensures valid examples are matching while invalid examples are failing? Currently, we don’t have any safeguards in place for potential changes to the schema

And can we ensure the CI is passing, at the moment its failing due to bad modifications made to the examples schema

core/src/main/resources/schemas/external_examples.schema.json
SaachiKaup: Correct Sufiyan, addressed

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: addressed - aligned with JSONSchema

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: addressed - aligned with JSONSchema

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: addressed - aligned with JSONSchema

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: addressed - aligned with JSONSchema

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: Addressed

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: Addressed

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: Addressed

core/src/main/resources/schemas/external_examples.schema.json
SaachiKaup: Have added the same, with 'examples' within the JSONSchema.

core/src/main/resources/schemas/external_examples.schema.json
SaachiKaup: Agreed - Resolved

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: Put it as nullable with allOf for the sibling ref

core/src/main/resources/schemas/external_example.yaml
SaachiKaup: Have added an example within JSONSchema for now and added some validation with format:duration (needed custom validation for YAML schema test)

core/src/test/kotlin/io/specmatic/mock/ExternalExampleOpenAPISchemaCorpusTest.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        val schemaValidation = FuzzyExampleJsonValidator.matches(jsonObject)
+        if (schemaValidation !is Result.Success) {
+            return schemaValidation
+        }
+
+        return validateFixtureDurations(jsonObject)
+    }
+
+    private fun validateFixtureDurations(example: JSONObjectValue): Result {
```
StarKhan6368: Can we avoid this ? it exists beacause we don't support format duration. However, if we add pattern to the string schema. This would eliminate the need for the current approach and allow us to validate the value during load.

core/src/main/resources/schemas/external_examples.schema.json
StarKhan6368: Not sure if `examples` will suffice, have you tried the schema with an actual example payload, does the autocomplete work with just examples ? examples also don't relay information on what the default value is, and what about descriptions ?

StarKhan6368 (changes requested):

PR #2591, opened 2026-06-19
Fix regression in handling of 4xx status codes beyond 400

joelrosario:
## Summary
- Prevent negative test generation from being filtered out when undeclared request variant 4xx responses are present alongside declared bad-request responses.
- Skip generative test mutation for undeclared request variants that have no examples, while still allowing example-backed generation.
- Add coverage for 415 responses declared via examples and for scenario generation behavior around undeclared request variants.

Fixes a regression caused by #2583


joelrosario:
Replaced by #2592 using the renamed branch codex/4xx-response-fix.

PR #2629, opened 2026-07-10
chore: bump specmatic reporter version

Sushant092003:
Update specmatic reporter version which fixes incorrect coverage percentages when combining multiple test reports.
Updated the HTML reporter library


yogeshnikam671 (approved):
LGTM

PR #2647, opened 2026-07-21
refactor: refactor BackwardCompatibilityCheckBaseCommand

### LINKED ISSUE #2644: Refactor Backward Compatibility Check code

yogeshnikam671:


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
```diff
@@ -0,0 +1,141 @@
+package application.backwardCompatibility
+
+import application.backwardCompatibility.BackwardCompatibilityCheckBaseCommand.ProcessedSpec
+import io.specmatic.core.IFeature
+import io.specmatic.core.Results
+import io.specmatic.core.log.logger
+
+internal class BackwardCompatibilityCheckLogger(
```
StarKhan6368: It appears to be a mixed design. Is this class intended solely for logging the outcomes of operations? Why are there methods that evaluate and return data used in the main flow?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
```diff
@@ -0,0 +1,141 @@
...
+        if (startWithNewLine) logger.log(newLine)
+        logger.log("-".repeat(20).prependIndent(ONE_INDENT))
+        logger.log("Verdict for spec $specFilePath:".prependIndent(ONE_INDENT))
+        logger.log("$ONE_INDENT$message")
+        logger.log("-".repeat(20).prependIndent(ONE_INDENT))
+        logger.log(newLine)
+    }
+
+    private fun logExampleValidationSummaryAndReturnResult(processedSpec: ProcessedSpec): Boolean {
+        val areExamplesInvalid = areExamplesValid(processedSpec.newer, "newer").not()
```
StarKhan6368: Dependency on the `areExamplesValid` lambda seems weird, why aren't we simply passing down validation outcome, instead of doing it here ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
```diff
@@ -0,0 +1,141 @@
...
+
+import application.backwardCompatibility.BackwardCompatibilityCheckBaseCommand.ProcessedSpec
+import io.specmatic.core.IFeature
+import io.specmatic.core.Results
+import io.specmatic.core.log.logger
+
+internal class BackwardCompatibilityCheckLogger(
+    private val baseBranch: String,
+    private val strictMode: Boolean,
+    private val hook: BackwardCompatibilityCheckHook?,
```
StarKhan6368: `strictMode` and `hook` appear to be necessary only at the end of the flow. Could we simply pass them down in that method? It seems there are no other use cases unless I've overlooked something.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -84,6 +84,10 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
         return isValidSpec(this)
     }
 
+    override fun File.isExternalisedExample(): Boolean {
+        return isFile && getParentExamplesDirectory(toPath()) != null
```
StarKhan6368: There is no confirmation that this is a example, it could be any file, not even JSON, located in a folder or a parent folder that ends with _examples or _tests.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -16,72 +16,39 @@ import io.specmatic.license.core.LicensedProduct
...
-    private val newLine = System.lineSeparator()
-    private var areLocalChangesStashed = false
-
     protected val effectiveRepoDir: String by lazy { options.repoDir ?: backwardCompConfig?.repoDirectory ?: "." }
     protected val gitCommand: GitCommand by lazy { SystemGit(workingDirectory = Paths.get(effectiveRepoDir).absolutePathString()) }
     protected val effectiveBaseBranch: String by lazy { options.baseBranch ?: backwardCompConfig?.baseBranch ?: gitCommand.currentRemoteBranch() }
     protected val effectiveTargetPath: String by lazy { options.targetPath ?: backwardCompConfig?.targetPath.orEmpty() }
     protected val effectiveStrictMode: Boolean by lazy { options.strictMode ?: backwardCompConfig?.strictMode ?: false }
 
+    val hook = ServiceLoader.load(BackwardCompatibilityCheckHook::class.java).firstOrNull()
```
StarKhan6368, line 40: Is there a reason this variable isn't private like the others in this class?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -16,72 +16,39 @@ import io.specmatic.license.core.LicensedProduct
...
+    private val backwardCompatibilityLogger by lazy {
+        BackwardCompatibilityCheckLogger(effectiveBaseBranch, effectiveStrictMode, hook, ::areExamplesValid)
+    }
+
+    private var areLocalChangesStashed = false
+
     abstract fun checkBackwardCompatibility(oldFeature: IFeature, newFeature: IFeature): BackwardCompatibilityCheckResult
     abstract fun File.isValidFileFormat(): Boolean
     abstract fun File.isValidSpec(): Boolean
+    abstract fun File.isExternalisedExample(): Boolean
```
StarKhan6368, line 49: `isExternalisedExample` may lead to downstream issues in other repositories. I assume there is either a PR addressing this, or we should implement a default otherwise

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -115,44 +82,58 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+        val changedFiles = gitCommand.getFilesChangedInCurrentBranch(
             effectiveBaseBranch
         ).filter {
             File(it).exists() && File(it).isValidFileFormat()
+        }.filter {
+            it.contains(Path(effectiveTargetPath).toString())
         }.toSet()
+
+        val externalisedExamples = changedFiles.filter { File(it).isExternalisedExample() }.toSet()
+        val specs = changedFiles.filter { it !in externalisedExamples && File(it).isValidSpec() }.toSet()
```
StarKhan6368, line 128: Building on the previous point, the criteria for identifying what constitutes an `externalExample` is weak, If that leads to false positives, we risk skipping valid specifications because of the filter here

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -115,44 +82,58 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
-        val filesReferringToChangedSchemaFiles = getSpecsReferringTo(filesChangedInCurrentBranch)
+        val specsReferringToChangedFiles = getSpecsReferringTo(changedFiles.specs + changedFiles.schemaFiles)
+            .filter { File(it).isValidSpec() }
+            .toSet()
 
-        val specificationsOfChangedExternalisedExamples =
-            getSpecsOfChangedExternalisedExamples(filesChangedInCurrentBranch)
+        val specsAssociatedWithChangedExternalisedExamples =
+            getSpecsOfChangedExternalisedExamples(changedFiles.externalisedExamples)
+                .filter { !File(it).exists() || File(it).isValidSpec() }
```
StarKhan6368, line 102: Why are we filtering them now (same above) ?
I assume we retain non-existing files for cases where example has changed but we don't have a matching specFile to associate it with, but it look like it gets skipped on actual check anyway ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -115,44 +82,58 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+        backwardCompatibilityLogger.logFilesToBeCheckedForBackwardCompatibility(
+            changedFiles.all,
+            specsReferringToChangedFiles,
+            specsAssociatedWithChangedExternalisedExamples,
             untrackedFiles
         )
 
-        val collectedFiles =
-            filesChangedInCurrentBranch + filesReferringToChangedSchemaFiles + specificationsOfChangedExternalisedExamples
+        val specsToCheck =
```
StarKhan6368, line 112: Is the behaviour change that only specFiles contribute to the final output expected  ? while the method is still named `getFilesToCheck` ? this was the reason for the tests diff i assume

StarKhan6368 (commented):
The changeset appears to have slightly deviated from a pure refactor. 
I hope we have ensured that there are no behavioral changes resulting from this.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: It has been non-private so did not change that and kept it as is to avoid issues anywhere else

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: I'll be raising a PR once this PR is approved and in pre-merge state.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: Yes this is expected. I've renamed the method back to getSpecsToCheck for clarity. Yes, this was the reason for the tests diff.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: Addressed by updating the isExternalisedExample check

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
yogeshnikam671: Addressed

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
yogeshnikam671: Addressed

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckLogger.kt
yogeshnikam671: Addressed

yogeshnikam671:
> The changeset appears to have slightly deviated from a pure refactor. I hope we have ensured that there are no behavioral changes resulting from this.

I am not sure where it deviated from a pure refactor. There may be gaps in the implementation that may have caused deviation like the isExternalisedExample thing that you pointed out but apart from it I do not see any deviation. Yes, there were some existing bugs that got fixed as part of the refactor but I don't think it counts as a deviation.

Please call out if I am missing something here.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -115,44 +80,58 @@ abstract class BackwardCompatibilityCheckBaseCommand(
...
+            it.contains(Path(effectiveTargetPath).toString())
         }.toSet()
+
+        val externalisedExamples = changedFiles.filter { File(it).isExternalisedExample() }.toSet()
+        val specs = changedFiles.filter { it !in externalisedExamples && File(it).isValidSpec() }.toSet()
+
+        return ChangedFiles(
+            specs = specs,
+            externalisedExamples = externalisedExamples,
+            schemaFiles = changedFiles - specs - externalisedExamples
```
StarKhan6368: I don't believe there's a reliable method to categorise files into schema and specification files. 

What are the  consequences of making an error in this categorisation? the internal check to determine if a file is a specification is weak imo

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
StarKhan6368: I see, it has been hoisted up, seems fine then we can let it be

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
yogeshnikam671: Yes.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt

StarKhan6368 (approved):
Looks good to me, thanks!

PR #2651, opened 2026-07-23
Fix/bcc operation request wiring

Sushant092003:
Make BCC operation requests protocol-aware

**Summary**
Backward-compatibility results previously assumed every scenario represented an OpenAPI operation. This PR introduces a protocol-aware operation-provider seam so each scenario can expose its correct API operation shape.

**Changes**

- Add ScenarioOperationProvider to derive an APIOperation from a Scenario.
- Keep OpenAPI as the default provider.
- Expose Scenario.toApiOperation() for consumers of scenario operation metadata.
- Remove OpenAPI-specific operation fields from the shared ScenarioDetailsForResult contract.
- Update backward-compatibility reporting to use scenario.toApiOperation().
- Update affected scenario, failure-report, contract-test, and externalised-example tests.


junit5-support/src/test/kotlin/io/specmatic/test/listeners/ContractExecutionListenerTest.kt
```diff
@@ -140,12 +155,11 @@ private fun testIdentifier(type: TestDescriptor.Type): TestIdentifier {
 }
 
 private class FakeScenario : ScenarioDetailsForResult {
```
StarKhan6368, line 142: Since `status`, `path`, and `method` serve no purpose here, why not remove them?

core/src/main/kotlin/io/specmatic/core/ScenarioOperationProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.specmatic.reporter.internal.dto.operation.APIOperation
+import io.specmatic.reporter.model.OpenAPIOperation
+
+interface ScenarioOperationProvider {
+    fun operationFor(scenario: Scenario): APIOperation
+}
+
+object OpenApiScenarioOperationProvider : ScenarioOperationProvider {
+    override fun operationFor(scenario: Scenario): APIOperation =
+        OpenAPIOperation(
```
StarKhan6368: Can we instead call `openAPIOperationFrom` or make it call this ? so we don't have duplication on how APIOperation for OAS is created from scenario

core/src/test/kotlin/io/specmatic/core/FailureReportTest.kt
```diff
@@ -7,12 +7,11 @@ import org.junit.jupiter.api.Test
...
 internal class FailureReportTest {
 
-    private class FakeScenario(override val status: Int) : ScenarioDetailsForResult {
+    private class FakeScenario(val status: Int) : ScenarioDetailsForResult {
         override val ignoreFailure: Boolean = false
         override val name: String = "fake scenario"
-        override val method: String = "GET"
-        override val path: String = "/fake"
-
+        val method: String = "GET"
```
StarKhan6368: Since  path, and method serve no purpose here, why not remove them?

core/src/test/kotlin/io/specmatic/core/ResultKtTest.kt
```diff
@@ -10,24 +10,23 @@ import java.util.function.Consumer
 
 internal class ResultKtTest {
     private val scenario = object : ScenarioDetailsForResult {
-        override val status: Int = 200
+        val status: Int = 200
```
StarKhan6368: Since status, path, and method serve no purpose here, why not remove them?

core/src/test/kotlin/io/specmatic/core/ResultKtTest.kt
```diff
@@ -10,24 +10,23 @@ import java.util.function.Consumer
...
+        val path: String = "/route"
         override fun testDescription(): String = "scenario description"
         override fun operationDescription(): String = "operation 1"
 
         override fun failureReportSubHeading(): String = "API: ${operationDescription()}"
     }
 
     private fun scenario(name: String, operationDescription: String) = object : ScenarioDetailsForResult {
-        override val status: Int = 200
+        val status: Int = 200
```
StarKhan6368: Since status, path, and method serve no purpose here, why not remove them?

core/src/test/kotlin/io/specmatic/test/ScenarioAsTestTest.kt
```diff
@@ -356,9 +356,7 @@ class ScenarioAsTestTest {
             val scenarioInRecord = testResultRecord.scenarioResult?.scenario as Scenario
 
             assertThat(updatedScenario.status).isEqualTo(422)
-            assertThat(updatedScenario.isNegative).isTrue
-            assertThat(updatedScenario.httpResponsePattern.headersPattern.contentType).isEqualTo("application/xml")
-            assertThat(scenarioInRecord.httpResponsePattern.headersPattern.contentType).isEqualTo("application/xml")
```
StarKhan6368, line 361: It appears that we have removed some assertions and added a different one. Is there a specific reason for this change?

core/src/test/kotlin/io/specmatic/core/ScenarioTest.kt
```diff
@@ -35,6 +36,30 @@ import org.assertj.core.api.Assertions.assertThat
 
 class ScenarioTest {
 
+    @Test
+    fun `operation is derived through the scenario operation provider`() {
```
StarKhan6368, line 41: It would be good to have a test with a custom `ScenarioOperationProvider` to ensure that we are delegating the call rather than always returning from or using `OpenApiScenarioOperationProvider`

StarKhan6368 (changes requested):

core/src/test/kotlin/io/specmatic/test/ScenarioAsTestTest.kt
```diff
@@ -320,7 +320,9 @@ class ScenarioAsTestTest {
         val scenarioInRecord = testResultRecord.scenarioResult?.scenario as Scenario
 
         assertThat(updatedScenario.httpResponsePattern.headersPattern.contentType).isEqualTo("application/xml")
```
StarKhan6368, line 322: Not a blocker but throughout this file, the assertions are now duplicated, `scenarioInRecord.responseContentType` is same as  `scenarioInRecord.httpResponsePattern.headersPattern.contentType`

StarKhan6368 (approved):
Looks good to me, please resolve the duplicate assertion before merge
And update the changelog if applicable

PR #2652, opened 2026-07-23
feat(#2645): add support for backward compatibility check to validate the changed externalised examples of compatible specs

### LINKED ISSUE #2645: Add Externalised Examples support in BCC
Acceptance Criteria:

* BCC should validate the changed externalised examples
* BCC should validate all the externalised examples whose spec/s has changed.

yogeshnikam671:
- Validate the changed externalised examples of backward compatible specs


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -156,6 +160,52 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
         return feature.loadExternalisedExamplesAndListUnloadableExamples().second
     }
 
+    override fun validateExamples(feature: IFeature, paths: Set<Path>): List<ExampleValidationResult> {
```
StarKhan6368, line 163: Since each example is validated in isolation, this will bypass duplicate example detection. I assume that should also be part of bcc  ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -156,6 +160,52 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+        feature as Feature
+        return paths.map { path ->
+            ExampleValidationResult(path, exampleValidationModule.validateExample(feature, path.toFile()))
+        }
+    }
+
+    override fun exampleValidationReportRecords(
```
StarKhan6368, line 170: Has there been any discussion regarding the inclusion of example validation in the BCC Ctrf Report? If so, what are the considerations for the source and affected sources? Currently, the source will be the file where the error occurred (refed file or specification itself), and that there may be multiple sources, as unlike BCC, in examples there can be more than one issue.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -156,6 +160,52 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+            contractPath = contractPath
+        )
+        else -> this
+    }
+
+    private fun scenarioFor(feature: Feature, path: Path): Scenario? =
+        ExampleFromFile.fromFile(path.toFile(), strictMode = false).realise(
+            hasValue = { example, _ ->
+                feature.scenarios.firstOrNull {
+                    it.matchesPathStructureAndMethod(example.request) && it.matchesStatusAndContentType(example.response)
```
StarKhan6368: We should consider the `requestContentType` in this context, as well as the potential for an incorrect method for status 405 and an incorrect content type for status 415

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -156,6 +160,52 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+            ExampleValidationResult(path, exampleValidationModule.validateExample(feature, path.toFile()))
+        }
+    }
+
+    override fun exampleValidationReportRecords(
+        feature: IFeature,
+        exampleValidationResults: List<ExampleValidationResult>
+    ): List<CtrfBackwardCompatibilityRecord> {
+        feature as Feature
+        return exampleValidationResults.mapNotNull { exampleValidationResult ->
```
StarKhan6368, line 175: Due to `mapNotNull` there can be cases where the CTRF report won't show any failures but the exitCode will still be `1`, not sure what the right way would be to resolve this

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt
```diff
@@ -327,6 +367,29 @@ abstract class BackwardCompatibilityCheckBaseCommand(
         })
     }
 
+    private fun validateChangedExternalisedExamplesIfApplicable(
+        processedSpec: ProcessedSpec,
+        examplesToValidate: ExternalisedExamplesToValidate?
+    ): ProcessedSpec {
+        if (examplesToValidate == null) return processedSpec
+        if (processedSpec.isNewFile.not() && processedSpec.backwardCompatibilityResult.successExcludingIgnorableFailures().not()) {
```
StarKhan6368: The hook can override the final decision for a BCC check, even if it fails due to the absence of consumers. The overall result would still be passed, and this guard clause does not appear to consider that.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt
```diff
@@ -156,6 +160,52 @@ class BackwardCompatibilityCheckCommandV2(options: BackwardCompatibilityCheckOpt
...
+            scenarioFor(feature, exampleValidationResult.examplePath)?.let { scenario ->
+                OpenApiBackwardCompatibilityCheckRecord(
```
StarKhan6368, line 177: This will also mark the operation as `changed`, even though it’s the example that has been altered, and the specFile remains unmodified

StarKhan6368 (changes requested):
My primary concern is that if we have determined that example validation must be included in the BCC report, which is essential since the exit code would not align with the failures otherwise, leading downstream jobs to mistakenly assume success if they only reference the CTRF failures.

We should ensure that the report remains unaffected in any way. Additionally, we must verify that no assumptions were made in the implementation regarding unchanged operations not failing, as this could happen once we incorporate the example results.

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckCommandV2.kt

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckBaseCommand.kt

PR #2661, opened 2026-07-27
fix: render before fixture  results info in studio when before fixtures fails

vedubhat:
Reason to change: 
 Studio did not render before execution results info when before fixture execution failed. This PR ensures that even if before fixture fails it gets rendered in studio

### DIFF
--- a/junit5-support/src/main/kotlin/io/specmatic/test/reports/TestHooks.kt
+++ b/junit5-support/src/main/kotlin/io/specmatic/test/reports/TestHooks.kt
@@ -48,18 +48,22 @@ internal fun getTestName(testResult: TestResultRecord, httpLogMessage: HttpLogMe
 }
 
 internal fun List<TestReportListener>.onTestResult(testResultRecord: TestResultRecord, testHttpLogMessages: List<HttpLogMessage>) {
-    val httpLogMessages = testHttpLogMessages.filter { it.scenario == testResultRecord.scenarioResult?.scenario }
-    if (httpLogMessages.isEmpty()) return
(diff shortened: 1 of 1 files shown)


junit5-support/src/main/kotlin/io/specmatic/test/reports/TestHooks.kt
```diff
@@ -48,19 +48,30 @@ internal fun getTestName(testResult: TestResultRecord, httpLogMessage: HttpLogMe
 }
 
 internal fun List<TestReportListener>.onTestResult(testResultRecord: TestResultRecord, testHttpLogMessages: List<HttpLogMessage>) {
-    val httpLogMessages = testHttpLogMessages.filter { it.scenario == testResultRecord.scenarioResult?.scenario }
-    if (httpLogMessages.isEmpty()) return
-    val firstHttpLogMessage = httpLogMessages.first()
-    val lastHttpLogMessage = httpLogMessages.last()
-    val testExecutionResult = TestExecutionResult(
+    val scenario = testResultRecord.scenarioResult?.scenario ?: return
```
yogeshnikam671: This changes the existing logic. Earlier if testHttpLogMessages genuinely had an entry with null scenario and testResultRecord.scenarioResult?.scenario was null, we would have got some httpLogMessages.
That behaviour changes here.

yogeshnikam671 (changes requested):

yogeshnikam671:
LGTM. @joelrosario please have a look once.

yogeshnikam671 (approved):

PR #2662, opened 2026-07-29
fix: main schema should override any member schema having the same property name from main schema during allOf parsing

yogeshnikam671:
- Fixed OpenAPI `allOf` conversion producing duplicate keys for same property: required `field` + optional `field?`.
- Main-schema properties now override matching `allOf` member properties, including optionality and pattern.
- Preserved top-level `required` behavior for properties defined only in member schemas.
- Added regression test.

### DIFF
--- a/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
+++ b/core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
@@ -3016,28 +3016,18 @@ class OpenApiSpecification(
         propertiesEntry: Map<String, Pattern>,
         propertiesAcc: Map<String, Pattern>
     ): Map<String, Pattern> {
-        val updatedPropertiesAcc: Map<String, Pattern> =
-            propertiesEntry.entries.fold(propertiesAcc) { acc, propertyEntry ->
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
```diff
@@ -3040,6 +3040,14 @@ class OpenApiSpecification(
         return updatedPropertiesAcc
     }
 
+    private fun override(
```
StarKhan6368: Could we simply fix the `combine` method instead as it also uses `restrictivePatternBetween` which we can maintain, in my head it would simply fix the removal of optional keys when adding new ones in

Then we could then add matrix test in `YamlToPatternTests` for top-level (optional, required) with member (optional, required)

StarKhan6368 (approved):
Looks good to me, thanks!

core/src/main/kotlin/io/specmatic/conversions/OpenApiSpecification.kt
yogeshnikam671: Addressed

PR #2675, opened 2026-08-05
fix: fix bcc mcp tool in case of docker

### LINKED ISSUE #2639: fix OSS backward compatibility MCP tool.
When the Specmatic MCP server is run inside a Docker container, operations that accept file or repository paths fail when the MCP client provides absolute host paths. 
For example, on Windows, Codex passes a path such as C:\specmaticProjects\sampleProjects\specmatic-order-contracts, but this path does not exist inside the Linux container. 
Although the host directory can be mounted into the container (for example, at /workspace), the absolute Windows path is not automatically translated to its corresponding container path. 
As a result, operations such as backward compatibility checks fail because the repository cannot be located within the container.

vedubhat:
Reason to change: 
BCC mcp tool was failing when use with docker. This PR ensures that the failure is handled by returning a docker bcc command as response by MCP as fallback.

### DIFF
--- a/application/src/main/kotlin/application/mcp/server/DefaultSpecmaticMcpToolProvider.kt
+++ b/application/src/main/kotlin/application/mcp/server/DefaultSpecmaticMcpToolProvider.kt
@@ -70,7 +70,7 @@ class DefaultSpecmaticMcpToolProvider(
     fun getBCCTool(): McpServerFeatures.SyncToolSpecification {
         return tool(
             name = "backward_compatibility_check",
-            description = "Check for breaking changes in OpenAPI specifications",
+            description = "Check for breaking changes in OpenAPI specifications. If the execution fails then it should return a fallback docker command and should not keeping on trying the bcc mcp tool again and again.",
             inputSchema = toolSchema(
                 properties = mapOf(
                     "targetPath" to stringProperty("File or folder path to analyze"),
--- a/application/src/main/kotlin/application/mcp/server/DefaultSpecmaticMcpToolProvider.kt
+++ b/application/src/main/kotlin/application/mcp/server/DefaultSpecmaticMcpToolProvider.kt
@@ -70,7 +70,7 @@ class DefaultSpecmaticMcpToolProvider(
     fun getBCCTool(): McpServerFeatures.SyncToolSpecification {
         return tool(
             name = "backward_compatibility_check",
-            description = "Check for breaking changes in OpenAPI specifications. If the execution fails then it should return a fallback docker command and should not keeping on trying the bcc mcp tool again and again.",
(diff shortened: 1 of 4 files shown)


application/src/main/kotlin/application/mcp/server/tools/BackwardCompatibilityTool.kt
```diff
@@ -11,14 +13,28 @@ data class BackwardCompatArgs(
...
+    val title: String = "Specmatic Backward Compatibility Check",
+    val message: String,
+    val dockerCommand: String,
+    val suggestion: String
+)
+
 class BackwardCompatibilityTool {
 
     internal fun runBackwardCompatibilityCheck(args: BackwardCompatArgs): String {
+        if(args.repoDir != null) {
```
yogeshnikam671: This whole check can be wrapped in a method named `isInvalidRepoDir`

application/src/main/kotlin/application/mcp/server/tools/BackwardCompatibilityTool.kt
```diff
@@ -11,14 +13,28 @@ data class BackwardCompatArgs(
...
+            if (!repoDirFile.exists() || !repoDirFile.isDirectory) {
+                return getFallbackResponse(args)
+            }
+        }
         val command = BackwardCompatibilityCheckCommandV2()
         val argsList = mutableListOf<String>()
         args.targetPath?.let { argsList.add("--target-path"); argsList.add(it) }
         args.baseBranch?.let { argsList.add("--base-branch"); argsList.add(it) }
-        args.repoDir?.let { argsList.add("--repo-dir"); argsList.add(it) }
+        args.repoDir?.let { argsList.add("--repo-dir"); argsList.add(args.repoDir) }
```
yogeshnikam671: Unnecessary change. `argsList.add(it)` is same as that of `argsList.add(args.repoDir)`

application/src/main/kotlin/application/mcp/server/DefaultSpecmaticMcpToolProvider.kt
```diff
@@ -70,7 +70,7 @@ class DefaultSpecmaticMcpToolProvider(
     fun getBCCTool(): McpServerFeatures.SyncToolSpecification {
         return tool(
             name = "backward_compatibility_check",
-            description = "Check for breaking changes in OpenAPI specifications",
+            description = "Check for breaking changes in OpenAPI specifications. If the execution fails then it should return a fallback docker command and should not keeping on trying the bcc mcp tool again and again.",
```
yogeshnikam671: Use the following description -

Checks for breaking changes between two OpenAPI specs.

Call at most once. Do not retry — failures are environmental, not transient.

On failure, returns a fallback docker command for the user to run locally. 
This is the final answer, not an error. Do not attempt the check via 
shell, other tools, or code execution. Just relay the command to the user.

yogeshnikam671 (changes requested):

yogeshnikam671 (approved):
LGTM

yogeshnikam671:
Update the changelog and then merge.

PR #2695, opened 2026-08-14
Fix/bcc verdict for hook

Sushant092003:
**What**:

Introduced a structured `CompatibilityResult` model for backward compatibility checks. Passed and failed results remain compatible through companion constants, while unknown results now carry a specific `UnknownReason`.

Updated Core production code and tests to use the new result structure.

**Why**:

The previous `UNKNOWN` result did not indicate why the compatibility check could not produce a definitive result. This made it difficult for consumers to distinguish cases such as an unavailable license, unsupported protocol, missing operation requests, or an Insights check failure.

**How**:

- Replaced the compatibility result enum with a sealed result model.
- Added `UnknownReason` values for supported unknown-result scenarios.
- Preserved existing `PASSED` and `FAILED` usage through companion constants.
- Updated compatibility report generation and default hook behavior.
- Updated affected Core tests.

### DIFF
--- a/core/src/main/resources/schemas/external_example.yaml
+++ b/core/src/main/resources/schemas/external_example.yaml
@@ -125,32 +125,6 @@ components:
         multipart-formdata:
           type: array
           items:
-            $ref: "#/components/schemas/MultiPartFormData"
-      required:
-        - path
-        - method
-
-    InlineRequestSchema:
-      type: object
-      additionalProperties: false
-      properties:
-        path:
-          type: string
-        method:
-          type: string
-        query:
-          type: object
-        headers:
-          type: object
-        body:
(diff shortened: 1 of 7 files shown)


core/src/main/resources/schemas/external_example.yaml
```diff
@@ -104,6 +104,32 @@ components:
       required:
         - method
 
+    InlineRequestSchema:
```
StarKhan6368, line 107: This change seems to have slipped in, can you please revert this so PR is clean

StarKhan6368 (changes requested):

StarKhan6368 (approved):
Looks good to me, thanks!

PR #2697, opened 2026-08-17
chore: update specmatic reporter version to latest which now adds column filter to html report

vedubhat:
- Bump specmatic-reporter version which now allows filtering based on the column in operation table in html report


yogeshnikam671 (approved):
LGTM

PR #2705, opened 2026-08-25
feat: wire CLI metadata into report generation

ketan:
This pull request introduces support for insights reporting options across the main Specmatic CLI commands (`stub`, `test`, and backward compatibility checks), enabling users to configure and validate reporting-related settings directly via command-line arguments. The changes also ensure that these options are properly integrated into the application's execution flow and tested accordingly.

### DIFF
--- a/application/src/main/kotlin/application/StubCommand.kt
+++ b/application/src/main/kotlin/application/StubCommand.kt
@@ -250,7 +250,6 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
             startServer()
 
             if (httpStub != null) {
-                ReportTracker.instance
                 if (registerShutdownHook) addShutdownHook()
 
                 val configuredHotReload = configuredHotReload()
--- a/application/src/main/kotlin/application/StubCommand.kt
+++ b/application/src/main/kotlin/application/StubCommand.kt
@@ -13,6 +13,9 @@ import io.specmatic.core.loadSpecmaticConfigOrNull
 import io.specmatic.core.toIncomingMtlsRegistryForStub
 import io.specmatic.core.utilities.Flags.Companion.SPECMATIC_BASE_URL
 import io.specmatic.license.core.cli.Category
+import io.specmatic.reporter.commands.GitReportDefaultValueProvider
+import io.specmatic.reporter.commands.InsightsReportOptions
+import io.specmatic.reporter.ReportTracker
 import io.specmatic.mock.ScenarioStub
 import io.specmatic.stub.HttpClientFactory
 import io.specmatic.stub.HttpStub
@@ -40,7 +43,8 @@ import kotlin.time.toDuration
     name = "mock",
     aliases = ["virtualize", "stub"],
     mixinStandardHelpOptions = true,
-    description = ["Start a mock server with contract"]
+    description = ["Start a mock server with contract"],
+    defaultValueProvider = GitReportDefaultValueProvider::class,
 )
 @Category("Specmatic core")
 class StubCommand(
@@ -145,6 +149,9 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
     @Option(names = ["--lenient"], description = ["Parse the OpenAPI Specification with leniency"], required = false)
     var lenientMode: Boolean? = null
 
+    @field:ArgGroup(exclusive = false, heading = "%nInsights reporting options:%n")
+    val insightsReportOptions = InsightsReportOptions()
+
     private var contractSources: List<ContractPathData> = emptyList()
 
     var specmaticConfigPath: String? = null
@@ -193,6 +200,7 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
     }
 
     override fun call(): Int {
+        insightsReportOptions.validate()
         configureLogging(
             LoggingFromOpts(
                 debug = verbose,
@@ -242,6 +250,7 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
             startServer()
 
             if (httpStub != null) {
+                ReportTracker.instance
                 if (registerShutdownHook) addShutdownHook()
 
                 val configuredHotReload = configuredHotReload()
@@ -322,7 +331,8 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
             gracefulRestartTimeoutInMs = configuredGracefulTimeout(),
             specToBaseUrlMap = contractSources.specToBaseUrlMap(),
             listeners = listeners,
-            requestHandlers = requestHandlers
+            requestHandlers = requestHandlers,
+            insightsReportOptions = insightsReportOptions,
         )
 
         LogTail.storeSnapshot()
--- a/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckOptions.kt
+++ b/application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckOptions.kt
@@ -1,8 +1,13 @@
 package application.backwardCompatibility
 
+import io.specmatic.reporter.commands.InsightsReportOptions
+import picocli.CommandLine.Mixin
 import picocli.CommandLine.Option
 
 class BackwardCompatibilityCheckOptions {
+    @field:Mixin
+    val insightsReportOptions = InsightsReportOptions()
+
     @Option(
         names = ["--base-branch"],
         description = ["Base branch to compare the changes against", "Default value is the local origin HEAD of the current branch"],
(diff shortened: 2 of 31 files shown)


application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckOptions.kt
```diff
@@ -1,8 +1,13 @@
 package application.backwardCompatibility
 
+import io.specmatic.reporter.commands.InsightsReportOptions
+import picocli.CommandLine.Mixin
 import picocli.CommandLine.Option
 
 class BackwardCompatibilityCheckOptions {
+    @field:Mixin
+    val insightsReportOptions = InsightsReportOptions()
```
StarKhan6368, line 9: It appears that the options here are not used ? in fact, we seem to be lacking the auto report-sending functionality for BCC generated reports

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
```diff
@@ -1186,7 +1189,9 @@ class HttpStub(
                 endTime = Instant.now().toEpochMilli(),
                 coverage = mockUsageReport.coverage,
                 absoluteCoverage = mockUsageReport.absoluteCoverage,
-                reportDir = File("${specmaticConfigInstance.getReportDirPath()}/stub")
+                reportDir = File("${specmaticConfigInstance.getReportDirPath()}/stub"),
+                reportType = RawReportType.MOCK,
+                insightsReportOptions = insightsReportOptions,
```
StarKhan6368, line 1194: Hot reload generates partial mock submissions. Each generation includes only the traffic recorded since the last reload. Is this by design ? Shouldn't we be only sending the final report for mocks instead ?

core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
```diff
@@ -41,6 +46,7 @@ object ReportGenerator {
             toolName = toolName
         )
 
+        ReportTracker.instance.send(report, reportType, insightsReportOptions)
```
StarKhan6368, line 49: When `--ci` is used, similar to implicit report publishing, it seems that the future is disregarded for CI runs, leaving the user without a clear indication of whether the report was successfully sent. If I remember correctly, the `send-report` command exited with the appropriate failure codes

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
StarKhan6368, line 1194: When no hot reload occurs, ReportTracker is first initialised from the mock’s JVM shutdown hook. Its constructor attempts to register another shutdown hook, which throws IllegalStateException("Shutdown in progress"). 

This prevents the final mock report from being generated and can leave the mock process hanging during Ctrl-C shutdown. The tracker must be initialised before shutdown instead of installing a nested hook.

StarKhan6368 (changes requested):
I don’t see automated tests for this reporting path. Could we add a few focused tests with the network call mocked ?
Also, can we confirm that Studio test/mock runs should send reports, and whether Studio’s BCC flow should send BCC reports as well ?

application/src/main/kotlin/application/backwardCompatibility/BackwardCompatibilityCheckOptions.kt
ketan: Fixed in b7a753d76. The BCC command now passes its parsed Insights options into BCC report generation, which submits the generated CTRF report as `BCC`.

core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
ketan: This path is intentionally asynchronous. Unlike `send-report`, command execution and its exit status remain about test/mock/BCC execution; the tracker flushes pending uploads during JVM shutdown rather than making an upload failure change the command result.

core/src/main/kotlin/io/specmatic/stub/HttpStub.kt
ketan: This is intentional: each hot-reload lifecycle is submitted before its in-memory traffic is discarded. Sending only the final lifecycle would lose traffic observed before a reload.

ketan: Fixed in b7a753d76. `StubCommand` initializes `ReportTracker` after startup and before registering its shutdown hook, so the tracker cannot attempt to install a nested hook while shutdown is already in progress.

ketan:
Follow-up on the Studio question: Studio’s CLI test and mock harnesses invoke the same JUnit/HTTP-stub report generators, so they submit through the normal reporting path. Studio’s interactive backward-compatibility endpoint only returns an in-memory UI response and does not generate a BCC CTRF report, so it does not submit one. Focused `BackwardCompatibilityCheckCommandV2Test` and `StubCommandTest` pass after the changes.

ketan:
> I don’t see automated tests for this reporting path. Could we add a few focused tests with the network call mocked ? Also, can we confirm that Studio test/mock runs should send reports, and whether Studio’s BCC flow should send BCC reports as well ?

The assumption is that you trust that the library does the right thing :)

junit5-support/src/main/kotlin/io/specmatic/test/ContractTestSettings.kt
```diff
@@ -40,6 +41,7 @@ data class ContractTestSettings(
     val previousRunCoverageMetrics: Map<OpenAPIOperation, CtrfOperationMetrics> = emptyMap(),
     val timeoutInMilliSeconds: Long? = null,
     val otherArguments: DeprecatedArguments? = null,
+    val insightsReportOptions: InsightsReportOptions = InsightsReportOptions.sniff().also { it.validate() }
```
StarKhan6368, line 44: `api.kt`, `HttpStub`, and `SpecmaticJUnitSupport` now validate `InsightsReportOptions.sniff()` by default. In CI, `sniff()` sets `ci=true` but does not infer `buildId`/`repoId`, so normal programmatic mocks and JUnit users can fail before execution ?

Can users provide a buildId when running programmatically? What if they choose not to publish reports during CI ? or disable shipping, validate doesn't seem to care ?

application/src/main/kotlin/application/StubCommand.kt
```diff
@@ -242,6 +250,7 @@ https://docs.specmatic.io/documentation/contract_tests.html#supported-filters--o
             startServer()
 
             if (httpStub != null) {
+                ReportTracker.instance
```
StarKhan6368: I’m not sure how likely this is in practice, but there still seems to be a shutdown race, the tracker and mock hooks run concurrently, so the tracker executor could close before `HttpStub.close()` submits the final report. That could cause `RejectedExecutionException`, lose the final report, and prevent `server.stop()` from running ?

application/src/main/kotlin/application/StubCommand.kt
```diff
@@ -40,7 +43,8 @@ import kotlin.time.toDuration
     name = "mock",
     aliases = ["virtualize", "stub"],
     mixinStandardHelpOptions = true,
-    description = ["Start a mock server with contract"]
+    description = ["Start a mock server with contract"],
+    defaultValueProvider = GitReportDefaultValueProvider::class,
```
StarKhan6368, line 47: The defaults seem out of sync, `InsightsReportOptions.sniff()` infers `ci` and `repoId` as well, but `GitReportDefaultValueProvider` only maps repo/branch fields into CLI defaults. So CLI and programmatic callers can end up with different inferred option values ?

junit5-support/src/main/kotlin/io/specmatic/test/ContractTestSettings.kt
StarKhan6368, line 44: `ReportTracker` calls `useDefaults()` before checking `shipDisabled`, and `useDefaults()` turns `CI=true` into `ci=true`. This can force CI submission even when the caller did not request it ?

StarKhan6368 (changes requested):

core/src/main/kotlin/io/specmatic/core/report/ReportGenerator.kt
StarKhan6368: Understood, my point was that the feedback is lost. Once I follow this process, I cannot ensure that a passing CI indicates the command was successful and that the results were sent to Insights as expected, i can only guarantee that the test didn't fail, or am i wrong about that ?

StarKhan6368:
> > I don’t see automated tests for this reporting path. Could we add a few focused tests with the network call mocked ? Also, can we confirm that Studio test/mock runs should send reports, and whether Studio’s BCC flow should send BCC reports as well ?
> 
> The assumption is that you trust that the library does the right thing :)

That makes sense. My point wasn’t to test the report-sending behavior itself, but rather the code path/wiring that invokes the library calls

StarKhan6368 (commented):
Looks good to me, thanks!

PR #2722, opened 2026-09-07
Add external mock handler extensibility via SPI

StarKhan6368:
**What**:
- Add per-spec mock mode resolution.
- Extract the existing `HttpExpectations` behavior into the default handler.
- Load mock mode handlers through `ServiceLoader`, with the core handler as fallback.
- Add an SPI for externally provided mock handlers and OpenAPI-backed routing coverage.

**Why**:

This change introduces an extension point for external mock implementations, such as stateful mocks.
> There is technically no behavioral change to core `HttpStub` usage: standard mocks continue using the existing matching, expectation, transient stub, token removal, and hook behavior through the default handler.

**How**:

- Resolve mock mode independently for each specification.
- Route requests and expectations to an external handler only when that mode is provided.
- Preserve the existing implementation as the default core handler.
- Add regression coverage for SPI loading and routing.

**Checklist**:

- [x] Unit Tests
- [ ] Build passing locally
- [ ] Sonar Quality Gate
- [ ] Documentation added/updated — N/A
- [ ] Sample Project added/updated — N/A
- [ ] Demo video — N/A
- [ ] Article on Website — N/A
- [ ] Roadmap updated — N/A
- [ ] Conference Talk — N/A

### DIFF
--- a/core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
+++ b/core/src/main/kotlin/io/specmatic/core/SpecmaticConfig.kt
@@ -5,6 +5,7 @@ import io.specmatic.core.azure.AzureAPI
 import io.specmatic.core.config.BackwardCompatibilityConfig
 import io.specmatic.core.config.LoggingConfiguration
 import io.specmatic.core.config.McpConfiguration
+import io.specmatic.core.config.MockMode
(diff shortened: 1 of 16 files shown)


yogeshnikam671 (approved):
LGTM

PR #2739, opened 2026-09-16
feat(config): support polymorphic async server entries

yogeshnikam671:
### **Why**:
AsyncAPI configurations in `specmatic.yaml` need to support both inline servers and `$serverRef` references. Support inline servers directly or under a `server:` wrapper and support $serverRef references under a `server:` wrapper.
Previously, the v3 schema did not permit these polymorphic shapes, and test configs were not delegating to the protocol validation SPI.

### **What**:
- Updated v3 JSON Schema to allow direct inline servers, wrapped servers, and `$serverRef` references.
- Connected `AsyncApiTestConfig` to the protocol validation SPI for per-spec validation.
- Added tests for schema validation and SPI invocation.

### **How**:
- Added `AsyncServerConfig`, `AsyncServerWrapper`, and `AsyncServerEntry` to `config-v3-resolved.schema.json`.
- Implemented `validateForSpecFile` in `AsyncApiRunOptions` to trigger `validateProtocolConfigForSpecFile`.
- Replaced rigid single-schema test cases in `ConfigSchemaValidationTest` with tests covering wrapped entries and references.


core/src/main/resources/config-validation/config-v3-resolved.schema.json
```diff
@@ -2009,6 +2009,82 @@
         }
       }
     },
+    "AsyncServerConfig": {
+      "title": "AsyncAPI server configuration",
+      "description": "An inline server connection or a local AsyncAPI server reference.",
+      "oneOf": [
```
StarKhan6368: Consider using `if`, `then`, and `else` instead of `oneOf` constraints when there is a clear discriminator field or value available, as this can lead to more cleaner error messages from the JSON Schema validator

core/src/main/resources/config-validation/config-v3-resolved.schema.json
```diff
@@ -2009,6 +2009,82 @@
         }
       }
     },
+    "AsyncServerConfig": {
```
StarKhan6368, line 2012: Consider using if, then, and else instead of oneOf constraints when there is a clear discriminator field or value available, as this can lead to more informative error messages from the JSON Schema validator.

core/src/test/kotlin/io/specmatic/core/config/validation/ConfigSchemaValidationTest.kt
```diff
@@ -378,6 +378,80 @@ class ConfigSchemaValidationTest {
...
+                                type: test
+                                servers:
+                                  - host: localhost:9092
+                                    protocol: kafka
+                                    client:
+                                      consumer: invalid
+                        """.trimIndent() to "servers/0/client/consumer",
+                    ).forEach { (config, invalidPath) ->
+                        val errors = schema(SpecmaticConfigVersion.VERSION_3, config)
+                        assertThat(errors + errors.flatMap { it.details })
```
StarKhan6368: The aggregate anyMatch assertion weakens coverage and appears to be compensating for the generic oneOf error. Once the schema uses conditionals, please restore the exact invalid type and invalid client.consumer tests, including their expected paths/metadata.

core/src/main/resources/config-validation/config-v3-resolved.schema.json
```diff
@@ -2009,6 +2009,82 @@
...
+      "required": [
+        "server"
+      ],
+      "properties": {
+        "server": {
+          "$ref": "#/definitions/AsyncServerConfig"
+        }
+      }
+    },
+    "AsyncServerEntry": {
```
StarKhan6368, line 2044: The PR description says $serverRef is supported directly or under server, but the current schema only accepts $serverRef inside the wrapper. Is this a miss on the PR description or in the schema ?

StarKhan6368 (changes requested):

core/src/main/resources/config-validation/config-v3-resolved.schema.json
yogeshnikam671: The PR description is wrong. I updated it.

core/src/main/resources/config-validation/config-v3-resolved.schema.json
yogeshnikam671: Addressed.

core/src/test/kotlin/io/specmatic/core/config/validation/ConfigSchemaValidationTest.kt
yogeshnikam671: Addressed.

StarKhan6368 (approved):
Looks good to me, thanks!

